PR #2, opened 2020-03-27
faqs.md: Add reference to schedv2.py

xuiqzy:
I don't know if the text description incorporates any changes in the v2 scheduler or that must be added, but it might be good to at least reference scheduler version 2, as that is the default being used in new versions.


dae:
cheers

PR #3, opened 2020-03-31
card reps vs card counts

guillemps:
That description no longer represents the actual behaviour.


dae:
It is correct for the v1 scheduler, which is still the default :-) Would probably be better to mention both cases instead.

guillemps:
I see. I have been using sched2 for so long. I'll take this into account in the future to avoid overlooking again. Feel free to adapt expressions or grammar.

dae:
Thanks!

PR #4, opened 2020-04-13
add brazilian portuguese manual link

mizerablebr:
Finishing some pages translations, I think that it is time to add the link to Brazilian Portuguese manual translation.


dae:
Thanks!

PR #6, opened 2020-05-06
Documented the ANKI_BASE environment variable

evandrocoan:
I was just about to open a pull request adding it, but while looking into the code I found out it was already implemented. This variable is useful because I can set it globally once and stop calling anki from everywhere.

https://anki.tenderapp.com/discussions/ankidesktop/41245-unwanted-file-movingdeleting-behavior


dae:
Thanks, I've pushed a tweaked version of this.

PR #7, opened 2020-05-07
Suggest fix for AnkiDroid mishandling tall images

gittywithexcitement:
AnkiDroid does not properly shrink tall images that have a aspect ratio that is tall and skinny. Suggest a fix.


dae:
"exmaple", but otherwise looks good thanks.

PR #10, opened 2020-07-19
Update studying.md

noobj2:
“\\~” comes after them. So you could call the deck “-Vocab” to make
them appear first, and you could call the other deck “\~ 
this makes the text strikethrough. doesn't appear on the preview tho, just appears on https://docs.ankiweb.net/#/studying?id=display-order


dae:
cheers

PR #9, opened 2020-07-17
Update info to 2.1.28 and some syntax

guillem-palau-salva:
Some info was outdated, i.e. the changes were introduced before .28. Feel free to adapt the wording in the changes that are more narrative.

I hope at least I saved you some time.

Cheers

PR #11, opened 2020-07-21
fix headings in editing.md

ANH25:
A small nitpick: the recently merged [pull request](https://github.com/ankitects/anki-manual/pull/9) of @guillem-palau-salva results in `h1` instead of `h2` for section headers.
I'm working on an Arabic translation of the manual so I noticed that :)

### DIFF
--- a/editing.md
+++ b/editing.md
@@ -67,7 +67,7 @@ site. In particular:
     understand the concepts behind the acronyms, learning the acronyms
     will become a lot easier.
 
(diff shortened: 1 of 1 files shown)


guillem-palau-salva:
So attentive. I may oversaw that. I checked now and this is the only file where this issue occurs. Thank you

dae:
Thanks ANH25!

PR #16, opened 2020-08-13
update browsing.md

ANH25:
Just some small changes.

By the way, the paragraph about the `edited` and `changed` columns needs to be updated. I didn't touch it since I'm not sure of all operations these columns track beside the stated ones. I think it would be nice to add more examples like the ones you've given in the [forum](https://forums.ankiweb.net/t/rename-card-modified-in-browser/1993/2).


dae:
Thanks ANH :-)

PR #27, opened 2020-10-26
Fix broken link to DropboxWiki

hydrox24:
The Unofficial Dropbox Wiki is no longer active, so I've replaced the broken link with an archive.org link.

I've also used a reference-style link with a link title because the URL is now long.


dae:
Thank you! Not sure if our system supports that link style - let's give it a go.

dae:
Looks good!

PR #30, opened 2020-10-27
directly link to the specific section

johan456789:
Linking to FAQs is not very helpful, and it took me some searches to find out the specific section it is referring to. 

As FAQs are moved to another repo recently in June, I assume the link will not change for a while. So I think we should directly link to the section using URL.

I'm not familiar with docify, please edit my file if the syntax is wrong. I guess the markdown syntax works.


dae:
Thanks!

PR #35, opened 2020-11-03
Update styling.md

johan456789:
link to specific section

### DIFF
--- a/templates/styling.md
+++ b/templates/styling.md
@@ -223,7 +223,7 @@ web if you’d like to learn more.
 
 If your card templates are complex, it may be difficult to read the
 question and answer columns (called "Front" and "Back") in the [card
-list](browsing.md). The "browser appearance" option allows you to define a
+list](browsing.md#card-list). The "browser appearance" option allows you to define a
 custom template to be used only in the browser, so you can include only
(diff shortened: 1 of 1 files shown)


johan456789:
@dae I'm reading through the manual, and I find it lacking screenshots. Anki is a GUI program, but there's no screenshot to illustrate where the buttons are at all. I think this is a huge problem. I sometimes feel that this manual is for already experienced users.

Like in this case, where is this [Browser Appearance](https://docs.ankiweb.net/#/templates/styling?id=browser-appearance) setting? It's actually hidden in `Options` -> `Browser Appearance...`. Users have to find it out themselves. This is bad user experience. Either the path should be included or there should be a screenshot or GIF showing that.

And IMO it'd be great if there's a screenshot showing what the window looks like:

image

Is it a deliberate choice not to contain any screenshots? Or just that no one has make them before?

dae:
There are two main reasons:

- The manual was previously one big page, and it would have been expensive to load all images at once
- Screenshots take time to make, and are more cumbersome to update when an option is added or removed

The former is no longer an issue, and if you'd like to put in the time on this, I'll gladly accept PRs that add screenshots.

PR #38, opened 2020-11-08
use American spelling

johan456789:
I see more places where American spelling is used. We should make it consistent.

### DIFF
--- a/browsing.md
+++ b/browsing.md
@@ -63,7 +63,7 @@ If you drag the mouse or hold ctrl or command to select multiple cards,
 the editor will be temporarily hidden. Various operations (such as
 changing the deck) can operate on multiple cards at once.
 
-The background colour will change depending on the card. Marked cards
+The background color will change depending on the card. Marked cards
 are a shade of purple. Suspended cards are a shade of yellow. For more
 information about marked and suspended cards, please see [editing and
 more](studying.md#editing-and-more).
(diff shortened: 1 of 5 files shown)


dae:
I'm not really sure how I feel about this. On the one hand, I try to use American variants in the code I write these days for consistency, and I gather there are more American English speakers out there. On the other hand, it's an international project, and my native tongue is actually Australian :-) Anyone want to chime in with their thoughts?

lovac42:
I'm used to seeing both colour and color, don't see why this should matter. And since dae is the one maintaining this, there might be future typos causing confusion with a mix of color and colour.

johan456789:
Mainly two reasons:
- Consistency: more visually pleasing, could be any English variant
- Search: Docsify's search seems to be string matching only

e.g I'll only find half of the results if I search either _color_ or _colour_. How would the user know s/he should try another spelling if s/he can't find something?

out-h


On the same note, Google doesn't index the whole Anki manual. @dae Could this be fixed?

image

dae:
You make a good point; I was thinking in terms of aesthetics and had not thought about ease of searching. That tips the scales for me, so I'll merge this in, and try to remember to use color moving forward.

The current docs generator does so using client-side JavaScript, which may be causing problems for Google. mdbook would solve this, but brings its own problems: https://github.com/ankitects/help-wanted/issues/7

dae:
(@lovac42 thanks for chiming in)

PR #42, opened 2020-11-25
Searching

RumovZ:
I've extended the searching chapter with a section about escaping in accordance with the revised parser.
Since you've merged the branch before I could reverse it, @dae, I'm assuming the `r:` prefix will stay and documented that as well. 😉

### DIFF
--- a/searching.md
+++ b/searching.md
@@ -101,8 +101,9 @@ you can search for each of them as a literal.
   `part:"after the colon"`.
 
 - `\`  
-  The backslash is used to deprive characters, including itself, of their special
-  meaning. This is called “escaping”. So `\\` must be used to match a single `\`.
+  The backslash is used to deprive the following character, possibly another `\`,
+  of its special meaning. This is called “escaping”. So `\\` must be used to match
+  a single `\`.
 
 - `"`, `*` and `_`  
   These three characters can only be matched literally if you escape them with `\`.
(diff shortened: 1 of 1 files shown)


dae:
Oh, I didn't actually notice you'd left that in ... which is impressive given I touched that specific code today when fixing the clippy lints. :-)

If it does stay, I think it probably should be given a longer name, as I presume it's not going to be something people are using that frequently - maybe 'raw:'?

RumovZ:
Ah, so it wasn't a deliberate decision. 😅
I can make a PR to remove it, not a big deal.

As for the name, `raw` was the second option. In the end, I chose `r` because I deemed it less likely to be used as a field name.

RumovZ:
Even though it now only applies to regex, I've kept the subsection about raw input since it may also be relevant for future functionality.
In fact, there already is another keyword that treats input like `re:`, `dupe:`. But apparently, it's not meant to be used manually and undocumented.

dae:
Thanks Rumo. Yeah, dupe: is not really intended to be used directly.

dae:
I've made some quick tweaks to these docs - if you disagree with any of the changes, please let me know

RumovZ:
Only two small notes:

1. The pointer to the Raw input subsection in the Regex section needs to be adjusted (it still says “above”).
2. I was trying to be escpecially assertive that quoting only works on a phrase level and not on a character level like escaping does because something like `left"("` won't cause an error but will probably lead to unexpected behaviour. Therefore, I would prefer to rephrase the sentence  
    >You can search for parentheses either by enclosing them in quotes [...]

dae:
Thanks, I've pushed out an update

PR #43, opened 2020-12-04
Corrected a likely typo in the filtered decks docs.

WraithGlade:
Compare

> such as when you need to **revise** for a test

to

> such as when you need to **review** for a test


dae:
Appreciate you taking the time to contribute, but in this case it's just a British English expression. :-)

johan456789:
Maybe not necessarily a typo, but it's better to use word senses that are common in both British and American English as Anki has global users.

Does ***review** for a test* mean basically the same thing as ***revise** for a test* in British English? If so, we should change it.

***revise** for a test* is confusing to me as an ESL speaker who has less exposure to British English.

PR #44, opened 2020-12-28
clarify retaining formatting when pasting

JCharante:
I ran across this section when skimming the docs. What is currently there does not match my experience, and posts from the forum [validate my experience (see below).

I did not know about the preference toggle until now, so I also a sentence to inform readers about its existence.

https://forums.ankiweb.net/t/paste-plain-text/4326

https://forums.ankiweb.net/t/ctrl-c-v-without-original-text-formatting/2921


dae:
Thanks!

PR #47, opened 2021-01-19
Add paragraph about search replacing

RumovZ:
Are the kbd tags fine? I would then update the rest of browsing.md accordingly as it makes it easier to distinguish key names and text input which are currently both styled as code.

### DIFF
--- a/browsing.md
+++ b/browsing.md
@@ -32,11 +32,11 @@ On Anki 2.1.39+, you can also hold down both <kbd>Ctrl</kbd> and
 <kbd>Shift</kbd> when clicking a search term to replace all occurences of the
 same kind of search with the new one.
 Let’s say you had previously typed in a complicated search expression like
-`deck:French (is:due or tag:important)`
-and now want to perform the same search for your Spanish deck. You can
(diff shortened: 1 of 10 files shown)


dae:
I think kbd is intended to look like an actual key, so the individual characters should be listed separately as in the demo on this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

Docsify supports it, and I'd be happy to accept such a change if you'd like to make it.

RumovZ:
Individual keys look way better, indeed. The tag nesting that they also suggest doesn't sound like a good idea, though, so I've decided to go with `<kbd>key1</kbd>+<kbd>key2</kbd>-click` and changed it for the whole manual.

dae:
Thanks!

RumovZ:
It turns out the kbd styling is terrible on Docsify. Can we add manual styling to fix this? 

grafik

dae:
Should be possible to inject some css into index.html - you can test locally by installing docsify-cli with npm/yarn, then running 'docsify serve'

PR #48, opened 2021-01-22
Overwrite negative left margin for kbd tags

RumovZ:
I just fixed the overlapping issue. The styling is ugly but even if I could do it better it would need to be adjusted with every theme change.
[This](https://github.com/jhildenbiddle/docsify-themeable/blob/7008137b8079bc267e66e653dedb18d8e417f605/src/scss/app/_content.scss#L260) is the line causing the issue. Maybe they'll fix it and we can revert this commit in the future.


dae:
Thanks Rumo!

In the long run I'm still leaning towards moving to mdbook for the manual, which should also indirectly solve this.

PR #49, opened 2021-02-04
Add name of Hermann Ebbinghaus to the Spaced Repetition history text

miluoshi:
Source: https://en.wikipedia.org/wiki/Hermann_Ebbinghaus


dae:
Thanks!

PR #50, opened 2021-03-07
Update preferences.md

Vogone:
Maybe I am heaving a bad day, then please ignore this change proposal, but after reading ten times, I concluded this is the way it actually behaves.


dae:
Thanks, this behaviour changed since that section was originally written.

PR #51, opened 2021-04-04
Add Anki 2.1 release updates to suspend/bury card behavior descriptions

missy-davies:
This would update the Studying section of the Manual to reflect the new release changes from [this update](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html) for behaviours of burying/suspending cards.

### DIFF
--- a/studying.md
+++ b/studying.md
@@ -189,7 +189,8 @@ With the old scheduler, if cards were in learning when they are buried,
 they are moved back to the new card queue or review queue prior to being
 buried.
 
-With the new scheduler, however, burying cards does not reset a card's learning steps.
+With the [new scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html),
+however, burying cards does not reset a card's learning steps.
 
 **Suspend Card / Note**: Hides a card or all of the note’s cards from review until they are
 manually unsuspended (by clicking the suspend button in the browser).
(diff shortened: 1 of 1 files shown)


missy-davies:
@dae I updated this as we [discussed](https://forums.ankiweb.net/t/a-thank-you-newcomer-contributions-to-the-manual/8885). I kept it small for now to see if I was going in the right direction. If this looks ok, then I'll continue working through the other Anki 2.1 scheduling changes.

studying.md
```diff
@@ -183,17 +183,25 @@ mainly left around for compatibility with older Anki versions.
 button on the [deck overview](studying.md#study-overview) screen.) This is useful if
 you cannot answer the card at the moment or you want to come back to it
 another time. Burying can also [happen automatically](studying.md#siblings-and-burying) for
-cards of the same note. If cards were in learning when they are buried,
+cards of the same note. 
+
+With the old scheduler, if cards were in learning when they are buried,
```
missy-davies, line 188: I noticed another section in the manual that kept info for both the old and new schedulers. Do we want to continue like this? Or focus just on relevant info for the latest version now that everyone has been prompted to update? 

If the later, then I can also update the documentation to remove references to the old verison.

dae: It will likely be quite a while until everyone has updated, as some users don't upgrade immediately, even if an update notification is sent out. So unfortunately we're probably best off mentioning both old and new behaviour for now.

Because people will often Google the part they're interested in, we can't rely on them to have read in chronological order, so it might also be worth turning one mention of the scheduler into a hyperlink in each section, eg 

With the [new scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html) however, burying cards ..

Otherwise looks good - thank you!

missy-davies: Thanks, makes sense! I have just now added in the links, so it should be good to go.

missy-davies: Thanks, that makes sense! I just added in the links, so it should be good to go now.

dae:
Thanks Missy!

PR #53, opened 2021-04-13
New scheduler updates filtered decks, relearning, buried cards

missy-davies:
This updates the manual to include some of the remaining changes from [the 2.1 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html) release, most notably information about filtered decks. 

This is part of the work for open issue: [update manual to reflect changes in 2.1 scheduler #1073](https://github.com/ankitects/anki/issues/1073)

### DIFF
--- a/filtered-decks.md
+++ b/filtered-decks.md
@@ -111,11 +111,13 @@ such as limiting to tags, finding cards forgotten a certain number of
 times, and so on. Please see the [searching](searching.md) section of the
 manual for more information on the different possibilities.
 
-Filtered decks can not pull in cards that are suspended, buried, or
-already in a different filtered deck. And unless you are using the
-experimental scheduler, they can not pull in cards that are in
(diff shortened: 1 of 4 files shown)


dae:
Thanks again Missy!

abdnh:
Doesn't the sentence "With the old scheduler, filtered decks cannot pull in cards that are
suspended, buried, or already in a different filtered deck" imply incorrectly that the new scheduler doesn't also have this limitation?

I was answering this [forum question](https://forums.ankiweb.net/t/browse-window-question/9313) and referred to the manual to make sure I'm getting things right.

dae:
Yep, good point - I've tweaked the wording to make this clearer. I decided to leave out the fact that cards can be suspended while they're in a filtered deck in the new scheduler, as that section is mainly focused on filtered deck creation.

PR #54, opened 2021-04-22
Notes mode

RumovZ:
Not sure how to best incorporate the mode explanations. Replacing every "cards" with "cards or notes" didn't seem practical.
Would it be possible to maintain/store different versions of the manual? By now, there are a lot of version references (“On Anki 2.1.35+...”) which can impede readibility and alienate new users who don't know which version they are on. For features like notes mode which have to be addressed in multiple places, it's hardly possible at all to accurately describe the behaviour for all versions. So it would be a great facilitation if the current manual could be archived and additions to the new repository wouldn't have to discuss legacy behaviour.


browsing.md
```diff
@@ -69,14 +81,22 @@ things. "Edited" tracks the last time changes were made to the _note_
...
+rows, the editor will be temporarily hidden. Various operations (such as
+changing the deck) can operate on multiple cards or notes at once, independent
+of the active mode. Therefore in Cards Mode, a note is considered to be selected
+if any of its cards is selected and in Notes Mode, a card is considered to be selected
+if its note is selected. If an operation only works for a single card (like
+showing card info) and Notes Mode is active, it's performed on the current note's
+card with index 1.
+
+The background color will change depending on the card and note.
+In Cards Mode, a row is assigned the first applicable of the following colors:
```
dae, line 95:
```suggestion
In Cards mode, a row is assigned the first applicable of the following colors:
```

Maybe a lowercase mode would be better in each of the references? Don't have strong feelings about it though.

browsing.md
```diff
@@ -69,14 +81,22 @@ things. "Edited" tracks the last time changes were made to the _note_
...
+The color of its flag if any; a shade of yellow if the card is suspended; a
+shade of purple if its note is marked. In Notes Mode, only the last criterion
+applies.
```
dae, line 98: This is a bit technical-sounding. How about something like this?

In Cards mode, Anki will change the color of the row depending on the state of the card. The first match will be used:

1. if flagged, use the flag color
2. if suspended, yellow
3. if marked, purple

In Notes mode, color is only applied to marked notes.

dae (commented):
Looks good! 

Re separate manual versions, here are some reasons why I haven't done that so far:

- I suspect many users are reaching the manual via a Google search and not the link inside the app. This may mean they would be reading a version that doesn't match the one they're using, and if the manual made no mention of changed functionality, it might be confusing to see mention of features that didn't exist or behaved differently.
- The mentions of added/changed functionality also serve as a bit of an advertisement for the newer versions - a user on an older version may see mention of some feature they think useful, and upgrade based on that.
- It would require some sort of version dropdown on the manual page, which is more work :-)

I feel like separate manual versions might work better on projects that have long release cycles?

browsing.md
```diff
@@ -3,10 +3,22 @@
...
-left, the *card list* on the top right, and the *current note* on the bottom
+left, the *card/note table* on the top right, and the *current note* on the bottom
 right. By positioning the mouse between two sections, it is possible to click
 and drag to expand one section and shrink the other.
 
+## Table Modes
+
+Anki 2.1.43+ offers two modes to show either cards or notes in the data table.
+You can change the current mode by clicking the switch in the top left corner
+right of the sidebar or pressing <kbd>Alt</kbd>+<kbd>T</kbd>. The switch also
```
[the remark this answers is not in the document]
RumovZ: Wouldn't that mean that it's inside the sidebar? Maybe "at the top between sidebar and searchbar"?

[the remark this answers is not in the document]
dae: "At the top, to the left of the search area"?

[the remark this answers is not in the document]
dae: I think people will hopefully be able to find it no matter what we choose :-)

RumovZ:
I understand these reasons, however, I would have thought that Docsify made this somewhat easier. For example on docs.rs, it's annoying that the search engine often takes you to an outdated version, but at least it's immediately apparent when this happens.

So if you don't think it's a problem that the version restricition of Notes mode (e.g.) is only mentioned once, I'll continue. For the browser columns, I want to add a table in a new Appendix section that lists the behaviour for both modes. Maybe the list of the menu actions could be moved there, too.

dae:
I'm still leaning towards [shifting the main manual](https://github.com/ankitects/anki/issues/977) over to mdbook. We could presumably implement a version selector for it as well, but it would be a bit of a project, so I'd be tempted to just proceed with the current approach for now. The mentioning of notes mode once works for me. Tables of features/columns/etc sound fine too, though as the manual is hyperlinked, perhaps it would be better to include them in the relevant section instead of hiding them away at the end of the manual like appendices usually are?

RumovZ:
Just pushed the commit. 🙂 
Let me know what you think. (Maybe I should drop the "Shows..." in every cell?)

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+The list of cards or notes can be sorted by every column except for the
+Question and Answer columns.
+
+
+| Column | Cards mode | Notes mode |
+|-|-|-|
+| Answer | Shows the back side of the card in one line with the question stripped. You can also choose a [custom format](templates/styling.md#browser-appearance) in the card type editor instead. | Shows the same as in Cards mode, only for the first card of the note. |
+| Card(s) | Shows the name of the card's template. | Shows the number of cards the note has. |
+| Card Modified | Shows the last time changes were made to the card (e.g., when you reviewed the card and the review history and interval were updated). | Shows the last time changes were made to one of the note's cards. |
+| Created | Shows the same as in Notes mode for the card's note. | Shows when the note was created. |
```
dae, line 15: In cases where both are the same, maybe we could we collapse the two columns into 1, and show the text once?

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| Deck | Shows the name of the deck the card is in. | Shows the number of different decks the note's cards are in, or the deck name if it's only one. |
```
dae, line 16: "or the deck name if all cards are in the same deck"?

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| Due | Shows the due date for cards in reviews or (re)learning and the position in the new card queue for new cards. The line is wrapped in brackets if the card is suspended or buried. Sorting is done by type and only then by date or position. | Shows the due date for the note's next due review or (re)learning card that is not suspended, buried or in a filtered deck. |
```
dae, line 17: ...or (re)learning, and the position...

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| (Avg.) Ease | Shows the card's ease unless it's a new card. | Shows the average ease for the note's cards that are not new. |
```
dae, line 18: "Shows the card's ease if it is not new"?

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| (Avg.) Interval | Shows the card's interval if the card is in reviews or relearning. | Shows the average interval for the note's cards that are in reviews or relearning. |
```
dae, line 19: in review x2

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| Lapses | Shows how often the card lapsed. | Shows how often the note's cards have lapsed altogether. |
```
dae, line 20: "...the card was answered Again" might be clearer to the layperson

"the total lapses for all cards of the note"?

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| Note | Shows the same as in Notes mode for the card's note. | Shows the name of the note's notetype. |
+| Note Modified | Shows the same as in Notes mode for the card's note. | Shows when the note (e.g. the content of a field) was last edited. |
+| Question | Shows the front side of the card in one line. You can also choose a [custom format](templates/styling.md#browser-appearance) in the card type editor instead. | Shows the same as in Cards mode, only for the first card of the note. |
+| Reviews | Shows how often the card has been reviewed. | Shows how often the note's cards have been reviewed altogether. |
```
dae, line 24: "the total review count for all cards of the note"?

appendix.md
```diff
@@ -0,0 +1,26 @@
...
+| Sort Field | Shows the same as in Notes mode for the card's note. | Anki allows you to choose one field from each type of note to be used for sorting and displayed in this column. You can change the sort field by clicking **Fields** in the current note section. |
```
dae, line 25: "...for sorting, and displaying in this column"

"current note section" → "editing area at the bottom" maybe?

dae:
Yep, "shows" does seem a bit redundant :-) I'd still lean towards including this in the browse section, so users can click on a section and read everything about it if they so wish - it's a bit hidden when in a separate appendix.

appendix.md
RumovZ: That would be great, but I don't think Markdown supports that?

appendix.md
RumovZ: I don't see it. 🤔

dae: sorry, I mean in reviews → in review needs to happen twice

appendix.md
dae: Ah, that's a shame :-(

appendix.md
RumovZ: This is copied from browsing.md where the editor is called "current note". Should I change it there as well? I find "editing area" clearer, too.

dae: Yep, please feel free to make any other adjustments you think would improve things

RumovZ:
I've read that @missy-davies intends to document the Forget action, so I've left it blank for now. 🙂 Set Due Date currently holds the text from Reschedule, but I believe we need separate entries for the two actions? Reposition hasn't changed to my knowledge, but should we move the lengthy text into its own section like e.g. Find and Replace?
Also, I think some more version restrictions need to be added. Maybe it would be a good idea to do that via footnotes.

browsing.md
```diff
@@ -102,29 +127,61 @@ box. Note that this will not display any type answer fields on your
...
+| Forget |  |
+| Reposition | Change the order new cards will appear in. You can find out the existing positions by enabling the _due_ column, as described in the [table](#cardnote-table) section above. If you run the reposition command when multiple cards are selected, it will apply increasing numbers to each card in turn. By default the number increases by one for each card, but this can be adjusted by changing the "step" setting. The **Shift position of existing cards** option allows you to insert cards between currently existing ones, pushing the currently existing ones apart. For instance, if you have five cards and you want to move 3, 4, and 5 between 1 and 2, selecting this setting would cause the cards to end up in the order 1, 3, 4, 5, 2. By contrast, if you turn this option off, 1 and 2 will get the same position number (and it will thus be unpredictable which of the cards with the same number comes up first). |
```
dae, line 176:
```suggestion
| Reposition | Change the order new cards will appear in. You can find out the existing positions by enabling the _due_ column, as described in the [table](#cardnote-table) section above. If you run the reposition command when multiple cards are selected, it will apply increasing numbers to each card in turn. By default the number increases by one for each card, but this can be adjusted by changing the "step" setting. The **Shift position of existing cards** option allows you to insert cards between currently existing ones, pushing the currently existing ones apart. For instance, if you have five cards and you want to move 3, 4, and 5 between 1 and 2, selecting this setting would cause the cards to end up in the order 1, 3, 4, 5, 2. By contrast, if you turn this option off, 1 and 2 will get the same position number (and it will thus be unpredictable which of the cards with the same number comes up first). Please note that when enabled, any card with a higher position will be modified, and all of those changed cards will need to be sent the next time you sync.  ||
```

dae: Hmm, diff didn't work well there - I added:

Please note that when enabled, any card with a higher position will be modified, and all of those changed cards will need to be sent the next time you sync.  |

browsing.md
```diff
@@ -102,29 +127,61 @@ box. Note that this will not display any type answer fields on your
...
+| Toggle Suspend | [Suspend](studying.md#editing-and-more) or unsuspend all selected cards, depending on whether the current card is suspended or not. |
+| Flag | Toggle the flags of all selected cards. Whether a flag is added or removed depends on whether the current card has this flag. |
```
dae, line 178:
```suggestion
| Flag | Toggle the flags of all selected cards. Whether a flag is added or removed depends on whether the current card has the chosen flag. |
```

dae (commented):
This extra detail and links to appropriate sections is great Rumo, I imagine many people will find it useful.

browsing.md
RumovZ: One of the drawbacks of having to write ridiculously long lines...
Thanks for writing it out for me.

RumovZ:
So should I add an entry for Rescheduling pointing to Set Due Date?
After that, I'll add the doc for the new sidebar and then I can open the mdbook PR.

As for footnotes, Docsify doesn't support them but Mdbook does it beautifully. So maybe that's an option in the future to reduce the number of times we have to type out version numbers and macOS keys.

dae:
Do you mean for the old "Reschedule" action? Maybe just a sentence "This action was called Reschedule in older Anki versions" to the end of the Set Due Date one? No strong feelings though. And 👍  to trying out footnotes

RumovZ:
I leave it as it is then. 🙂 
Struggled a bit with documenting the new sidebar, so I'm sure there are improvements to be made.

browsing.md
```diff
@@ -38,92 +57,171 @@ and now want to perform the same search for your Urdu deck. You can
...
+and <kbd>Shift</kbd>-clicking. It also enables drag and drop to reorder decks and
+tags.
+
+Here is an example: Say you have the tags `Math`, `Calculus` and `Algebra`.
+Click on the `Calculus` tag, then <kbd>Ctrl</kbd>-click on the `Algebra` tag. Now both
+tags are selected, click and drag any of the two onto the `Math` tag to make them
+both children of this tag. Behind the curtains, Anki has renamed the two tags to
+`Math::Calculus` and `Math::Algebra` respectively and updated your notes accordingly.
+
+Another use case for selecting multiple items is searching: If you rightclick on
```
dae, line 72:
```suggestion
Another use case for selecting multiple items is searching: If you right-click on
```

browsing.md
```diff
@@ -38,92 +57,171 @@ and now want to perform the same search for your Urdu deck. You can
...
+a selection of items, you can choose **Search &gt; All/Any Selected Items**. This
+can be combined with keyboard modifiers as described in [Search Tool](#search-tool)
+to append the resulting search to the current search.
 
-To remove tags that are not used by any notes, use **Tools &gt; Check Database**
-from the main window.
+### Saved Searches
+
+If you regularly search for the same thing,
+you can save the current search by rightclicking the topmost item in the sidebar,
```
dae, line 80:
```suggestion
you can save the current search by right-clicking the topmost item in the sidebar,
```

browsing.md
```diff
@@ -38,92 +57,171 @@ and now want to perform the same search for your Urdu deck. You can
...
+choosing “Save Current Search” and typing in a name.
+
+### Editing Items
+
+You can delete or rename tags, decks and saved searches directly from the sidebar
+through the context menu or using the according keys (<kbd>Del</kbd> and <kbd>F2</kbd>
```
dae, line 86:
```suggestion
via the right-click menu, or by using a shortcut key (<kbd>Del</kbd> and <kbd>F2</kbd>
```

browsing.md
```diff
@@ -38,92 +57,171 @@ and now want to perform the same search for your Urdu deck. You can
...
-from the main window.
+### Saved Searches
+
+If you regularly search for the same thing,
+you can save the current search by rightclicking the topmost item in the sidebar,
+choosing “Save Current Search” and typing in a name.
+
+### Editing Items
+
+You can delete or rename tags, decks and saved searches directly from the sidebar
```
dae, line 85:
```suggestion
You can delete or rename tags, decks and saved searches directly from the sidebar,
```

browsing.md
```diff
@@ -38,92 +57,171 @@ and now want to perform the same search for your Urdu deck. You can
...
+*current*, card or note, which is usually the one that was last left- or rightclicked.
+In Cards Mode, again, the current note is the note of the current card and in Notes
+mode, the current card is the first card of the current note.
 
-## Current Note
+The background color will change depending on the card and note.
+In Cards mode, Anki will change the color of the row depending on the state of
+the card. The first match will be used: 1. if flagged, use the flag color; 2. if
+suspended, yellow; 3. if marked, purple. In Notes mode, color is only applied to
+marked notes.
```
dae, line 124: Maybe this would be a little easier to read as a separate list, instead of contained in the paragraph?

dae (commented):
Just some very minor comments :-)

RumovZ:
By the way, it says about Set Due Date:

> If you want to hide the history, you will need to export your notes as a text file, delete the notes, and then import the text file again, creating new notes.

Isn't that what Forget is for?

dae:
Well, neither are intended to reset history, but you're right that that comment would be more appropriately mentioned in the Forget row. :-)

RumovZ:
Ah okay, I've never used Forget or Set Due Date myself, almost proud of it. 😉
This PR can be merged then as far as I am concerned.

dae:
Thanks Rumo :-)

PR #55, opened 2021-04-24
Mdbook

RumovZ:
I tried splitting getting-started.md and it works reasonably well in my opinion. However, it also became apparent that this won't work for most other files. Splitting up all subsections automatically with a script was quite enlightening. Following are some thoughts and recap:

* Many files have a whole lot of subsections, let's take Studying as an example. I see two major issues with splitting them up:
  * Most subsections and especially the top-level section don't have a lot of text. As a reader, I don't like constantly turning pages while seeing so much of my screen space wasted.
  * Mdbook doesn't have collapsable sections as Docsify does, so every new file will take up vertical space in the sidebar. The sidebar is already almost one screen height long (a questionable unit of measurement, but you get my point). I feel that if it gets much longer, it will lose usefulness quickly.
* Not splitting on the other hand also doesn't work well. E.g. Getting Started/Key Concepts is already a nested page, but without a more detailed TOC, the user has still no idea what information he can find there without scrolling through the page.

In conclusion, I don't see this working without an additional tool like mdbook-toc.
I would (very scarcely) split up some of the pages so the user can roughly see from the sidebar which page contains the information he's looking for, like I did with Getting Started. Some of the shorter pages, like Media or Add-ons, can maybe be made a subsection on another page to make room in the sidebar. Generally, there is probably some room for improvement regarding the structure/hierarchy to make things easier to find. But when all that is done, the sidebar is well-formed and the user quickly finds the relevant page, mdbook-toc is still needed to ergonomically navigate these long pages.

I have to say that, while I see mdbook as generally superior to dosify, it seems, true to its name, more suitable for a book that is read from front to back than for a manual that is more typically used as a reference document.


dae:
As an experiment I've added mdbook-toc and updated the key concepts section (the heading level needed to be adjusted) - what do you think?

dae:
The sections on the left can be collapsed by the way:

```
[output.html.fold]
enable = true
```

Not sure whether it would be best to do that, or keep one sublevel, to make it faster to jump to a particular section.

RumovZ:
> The sections on the left can be collapsed by the way:

Oops, quite an oversight!
mdbook-toc is my next misconception: When reading it would add a toc "on the right" I thought it would add a sidebar or floating element. At the top of the page, it's a lot less useful.

So I have to reevaluate, sorry about that.
You said that we could link to the subsections from the sidebar without actually splitting the files, would that be hard to do?
Apart from that, it now seems reasonable to me to split up the sections, but in such a way that we don't end up with half-empty pages.

> Not sure whether it would be best to do that, or keep one sublevel, to make it faster to jump to a particular section.

At the current state, I feel like we'd lose more speed to scrolling than we'd save from not having to expand. But maybe we could restructure so we have less top-level sections and only second-level sections that are valuable when shown in the sidebar?

How shall I proceed?

dae:
I was hoping we could use anchors in SUMMARY.md like this:
```
  - [Adding a Notetype](editing.md#adding-a-note-type)
```

but unfortunately mdbook doesn't support that.

I'm not actually too bothered by having the links at the top, as it'll be in view when users click on a section on the left, and they can click on the title to jump back to the top. We could potentially inject some CSS to fix the TOC to the right of the screen when the screen width is big enough though:
Screen Shot 2021-04-25 at 5 00 02 pm

dae:
> At the current state, I feel like we'd lose more speed to scrolling than we'd save from not having to expand. But maybe we could restructure so we have less top-level sections and only second-level sections that are valuable when shown in the sidebar?

> How shall I proceed?

With the toc floated on the right, we'd have something similar to the current docsify layout, and hopefully it wouldn't require too much moving things around? We would need to update headings to the appropriate levels and insert a toc marker in each page where a toc was required. There may be some sections that would benefit from being rearranged/split up as well, but presumably we can retain the current structure for some/most of them?

RumovZ:
So I'll make sure the files work with mdbook-toc and we postpone restructuring files for the time being. I'll make a new PR once the pending edits are incorporated.

PR #56, opened 2021-04-25
Introduced

RumovZ:


searching.md
```diff
@@ -293,7 +293,7 @@ included even if their notes were added a long time ago.
 cards where the note text was added/edited in the last n days. Requires
 Anki 2.1.28+ / AnkiMobile 2.0.64+.
 
-## Recently answered
+## Recently answered (for the first time)
```
dae, line 296: (for the first time) only applies to introduced:, so perhaps it's a bit confusing to have this in the title?

searching.md
```diff
@@ -307,7 +307,14 @@ cards answered Again (1) over the last 7 days
 `rated:31:4`  
 cards answered Easy (4) in the last month
 
-For speed, rating searches are limited to 31 days.
+Rating searches had been limited to 31 days before version 2.1.39.
+On version 2.1.43+, you can also search for the very first review only:
```
dae, line 311: probably need a newline between those two sentences, since they're unrelated

searching.md
RumovZ: It was supposed to apply to both keywords, therefore the brackets. Does the new approach work for you?

dae:
Even better, thanks Rumo :-)

PR #57, opened 2021-04-28
Mdbook

RumovZ:
I'm at a loss with mdbook-toc, it only works for the one file you set it up for on the other branch. It can't be my installation therefore, book.toml is the same, and I've checked countless times that nothing's wrong with the marker or the heading levels.
What am I missing? I'm sure it's something obvious...


dae:
Look good!

The TOC issue seems to be an edge case caused by book.toml having Windows linefeeds in it. I added the marker= line originally to try to get the TOC to display, but that seems to have caused other problems. If you remove the marker= line and save the file with Unix line endings, `<!-- toc -->` should start working in the top section of a file, eg

```diff
diff --git a/src/searching.md b/src/searching.md
index 236bd14..26e57a8 100644
--- a/src/searching.md
+++ b/src/searching.md
@@ -3,6 +3,8 @@
 Anki's Browse screen and the Filtered Deck feature use a common method
 of searching for specific cards/notes.
 
+<!-- toc -->
+
 ## Simple searches
 
 When you type some text into the search box, Anki finds matching notes
```

RumovZ:
Oh dear. Thanks for soting it out.

Shall I look into fixing the toc to the top of the screen?

dae:
Sure, if have time that’d be great- the additional css option could
probably do what we need

RumovZ:
Unfortunately, I can't get it to work well. There are a few problems:

1. After taking into account the sidebar and page navigators (the ones for previous/next page), the available space is quite small, actually.
2. Navigators and/or sidebar may be hidden, which makes calculating position and width of the toc quite complex.
3. Even when taking into account the aforementioned variables, my calculation doesn't quite work out. (This may well be due to my inexpertness with CSS.)

On a different matter, I've added some styling for kbd tags, but I'm not perfectly happy with the colour scheme for the two light themes.

dae:
No problem, thanks for looking into it. I'll have a play around over the next few days and see what I can come up with.

dae:
(I've deployed this as-is to main to start with - those new tables are looking good, as is the kbd styling!)

dae:
Ok, here's a brief proof of concept on the Browsing page:

https://github.com/ankitects/anki-manual/tree/floating-toc

In order to get it to fit on screens that aren't as wide, I had to shift the left margin, meaning the text is no longer centred on the right side, and I had to turn off the forward/backward arrows. It could probably be improved a bit with some polish, but it still feels a bit crowded - I wonder whether what we currently have might be enough?

RumovZ:
The navigators overlaying the text look quite odd, the main content doesn't look good left-aligned when the sidebar is hidden and it only works for a specific screen width, anyway. I assume it would take a lot of time to address these and other issues.
So yes, I would agree that the status quo should be good enough. Maybe some time could instead be spend on optimising the structure of the manual.

dae:
Yep, I think you're right. The manual could certainly still do with some more attention, so if that's an area that interests you, further changes are welcome :-)

abdnh:
It's a shame that the multilingual search in mdBook is a bit limited compared to Docsify. I just moved the Arabic manual to mdBook and realized late that the search bar is useless now :(

Support is limited even for languages like Japanese. (Try searching for 日本語 in the intro page). Of course, that's not an issue for the English manual.

dae:
Ouch, I was not aware of that. It looks like it's using elasticlunr under the hook, and it may be possible to add support for Arabic?

https://github.com/MihaiValentin/lunr-languages

abdnh:
Yes, I looked into elasticlunr. Thanks for the link! I will have a try at porting their Arabic stemmer to elasticlunr.

PR #58, opened 2021-05-02
Offline Manual Creation with Pandoc

james-wash:
I wanted to have an offline version of the manual available as well as an online version, so if I want or need to I can read through the documentation without an internet connection. So, I've made various changes to make this happen.

All of the markdown files have been modified to end with a \newpage command, so when Pandoc is run to splice the Markdown files together it will automatically start a new page for each major section. I've added a lua script file to act as a Pandoc filter for this functionality to work properly. I've also added a txt file outlining the command to stitch the markdown files in proper order into a docx file with Pandoc, and I've also included the output docx and pdf files as well.


dae:
Hi James,

We've had a change to a new generator brewing in #57 that I've just released. It has a print button at the top that will generate a one-page file to print, which should largely remove the need for a separate solution - though I appreciate you taking the time to contribute it in any case! For adding page breaks for each major section, I imagine we may be able to accomplish that with a bit of css.

james-wash:
Understandable. I looked at it and it looks good - the only suggestion I have is that you find a way to include a table of contents as well, as the document is lengthy and users may have trouble finding what they want without it.

PR #59, opened 2021-05-02
Fix version references to 2.1.44+

RumovZ:
I realised 2.1.43 is the current release.


dae:
I've increased it one more, so we can get another small patch release out

PR #61, opened 2021-07-11
Updating deck options to 2.1.45

cqg2:
Just reorganized the old page a little bit, adding some extra information from here and there. Not familiar with github, please feel free to ignore the PR if necessary.

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -269,6 +269,16 @@ Allows you to apply a minimum limit to the above
 option. The default setting says that lapses should be reviewed one day
 later. The interval must be 1 day or more.
 
+#### Example.
+- Relearning Steps: 10m 1h
+- Minimum interval: 1d
+
+If you press "Again" on a card that you reviewd 30 days ago (interval 30):
+- Anki will show the card again in 10 minutes.
+- If you press 'Good', Anki will show the card again in 1 hour. 
+- If you press 'Good', Anki will show the card again the next day (interval 1)
+, in review mode. 
+
 ### Leech options
 Control the way Anki handles leeches. Please see the
 [leeches](leeches.md) section for more information.
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -15,6 +15,8 @@ at the bottom of the screen.
 - Clicking on 'More' > 'Option' while in review mode. 
 - Pressing "o" while in review mode. 
 
+## Managing Deck Options
+
 Anki allows you to share options between different decks, to make
 updating options in many decks at once easy. To do this, options are
 grouped into an 'options group'. By default, all newly created decks use
@@ -104,8 +106,12 @@ The delay between answering 'easy' on a learning
 card and seeing it in review mode for the first time.
 
 ### Insertion Order
-Controls whether Anki should add new cards into the deck
-randomly, or in order. When you change this option, Anki will re-sort
+Controls whether Anki should add new cards into the deck. The options are:
+
+- Sequential (older cards first)
+- Random
+
+When you change this option, Anki will re-sort
 the decks using the current option group. One caveat with random order
 mode: if you review many of your new cards and then add more new cards,
 the newly added material is statistically more likely to appear than the
@@ -118,6 +124,7 @@ shown in the order their card types appear in, so that siblings are
 introduced consistently - otherwise you could end up in a state where
 some notes had all their cards introduced and other notes had only one
 or two. Please see the "bury related" option below for more info.
+
 
 ## Lapses
 
@@ -145,7 +152,6 @@ would accomplish the same thing as the "Good" button. If you have 2 or
 more steps, "Easy" is not hidden, to allow you to graduate cards from
 the queue before all of their steps are finished.
 
-
 ### Minimum interval
 Allows you to apply a minimum limit to the above
 option. The default setting says that lapses should be reviewed one day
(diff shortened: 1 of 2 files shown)


src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
-
-Deck options are accessed by selecting a deck on the 'Decks' screen, and
-then clicking 'Options' at the bottom of the screen.
+Deck options are probably one of the most important things to understand in
+order to get the most out of Anki. These options control Anki Scheduler, 
+the core of the program, so if inappropriate adjustments are used you could
+have to review too many cards every day, getting easily frustrated, or 
+you might not see cards in a deck often enough, rendering Anki less 
+effective. So please, only change options that you fully understand.
+
```
dae, line 10: Perhaps a bit less wordy?

```suggestion
Deck options primarily control the way Anki schedules cards. It is recommended that you spend a few weeks with the defaults to get a feel for how Anki works before you start adjusting options. Please make sure you understand the options before changing them, as mistakes could reduce Anki's effectiveness.
```

src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
+Deck options are accessed by:
+- Clicking the gears icon on the 'Decks' screen.
```
dae, line 12:
```suggestion
- Clicking the gear icon on the 'Decks' screen.
```

src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
+- Selecting a deck on the 'Decks' screen, and then clicking 'Options' 
+at the bottom of the screen.
+- Clicking on 'More' > 'Option' while in review mode.
```
dae, line 15:
```suggestion
- Clicking on 'More' > 'Options' while in review mode. 
```

src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
-one deck but not other decks, click the gears icon in the top right and
-add a new options group.
+Anki have separate option groups. 
 
-Please only change options that you fully understand, as inappropriate
-adjustments may render Anki less effective.
+If you’d like to alter the settings on
+one deck but not other decks, click the arrow icon in the top right 
+of the Deck Options window and **add** a new preset. From there you can 
+also **clone, rename** or **delete** the actual preset, or even **save the current
```
dae, line 29:
```suggestion
also **clone**, **rename** or **delete** the actual preset, or even **save the current
```

dae: We should probably explain what each of these do.

src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
+number of times before the delay between repetitions can increase
+appreciably. If you are consistently learning 20 new cards a day, you
+can expect your daily reviews to be roughly about 200 cards/day. You can
+decrease the reviews required by introducing fewer new cards each day,
+or by turning off new card display until your review burden decreases.
+More than one Anki user has excitedly studied hundreds of new cards over
+their first few days of using the program, and then become overwhelmed
+by the reviews required.
+
+If using [2021 scheduler (v3)](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html),
```
dae, line 60:
```suggestion
If using [the 2021 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html),
```

src/deck-options.md
```diff
@@ -1,34 +1,96 @@
...
+please keep in mind that the new count is capped by the review count. 
+If you have limits of 200 reviews and 20 new cards, and 190 reviews are
+due, only 10 new cards will be introduced.
+
+###  Maximum reviews/day
```
dae, line 65: Switching between "Title Case" and "Not title case" in these headers

src/deck-options.md
```diff
@@ -65,61 +149,158 @@ introduced consistently - otherwise you could end up in a state where
...
-**Graduating interval** is the delay between answering 'Good' on a card
-with no steps left, and seeing the card again.
+When you forget a review card, it is said to have 'lapsed'. The default
+behaviour for lapsed reviews is to reset the interval to 1 (ie, make it
+due tomorrow), and put it in the learning queue for a refresher in 10
+minutes. This behaviour can be customized with the options listed below.
 
-**Easy interval** is the delay between answering 'easy' on a learning
-card and seeing it in review mode for the first time.
+### Relearning steps.
```
dae, line 160:
```suggestion
### Relearning Steps
```

dae (commented):
Thank you for contributing!

I've left some comments on the first half. Some other thoughts:

- I noticed that there's a fair bit of overlap with the info shown in the tooltips inside the deck options screen. I wonder if we're adding much value by doing that? It makes the manual harder to maintain, and users will be left trying to dig through the manual to find the new pieces of information buried in the duplicate text.

- Because it'll be a while until all users have updated, we may need to maintain the old text as well, perhaps in two different sections (old/new Anki). Probably easiest to just make the changes you want for now, and we can then copy+paste in the old text right at the end.

dae:
For future reference, GitHub lets you automatically apply suggested changes, so you don't need to make them manually

cqg2:
> Thank you for contributing!
> 
> I've left some comments on the first half. Some other thoughts:
> 
> * I noticed that there's a fair bit of overlap with the info shown in the tooltips inside the deck options screen. I wonder if we're adding much value by doing that? It makes the manual harder to maintain, and users will be left trying to dig through the manual to find the new pieces of information buried in the duplicate text. My thinking with the manual is that it should explain concepts in more detail when more detail is required, but not repeat itself for the sake of completeness.
> * Because it'll be a while until all users have updated, we may need to maintain the old text as well, perhaps in two different sections (old/new Anki). Probably easiest to just make the changes you want for now, and we can then copy+paste in the old text right at the end.

Yes, you're right, I simple felt that the new tooltips offer a clearer and better explanation than the manual, that's why I replaced the old text. But yes, having duplicated info probably makes no sense. I'll continue editing the document, but not being a native speaker, I'm not sure whether it will help, or rather the opposite, specially if someone else has to correct my mistakes. So again, please feel free to simply ignore the PR if it's not helpful!

dae:
You are very fluent - I didn't even notice you are not a native speaker :-)

There's still a lot of duplication with the options page:

- a fair bit of text still appears to have been copied verbatim from the decks screen - are you planning to remove it, or what is your plan there?
- listing out available options makes sense if you're planning to explain them in more detail, but I don't think it adds a lot of value if you're just listing the options, as the list can already be seen on the page

Also because this is turning into quite a few changes, it may be worth splitting it up into multiple pull requests, and focusing on one section at a time, instead of trying to tackle the entire chapter at once.

cqg2:
> You are very fluent - I didn't even notice you are not a native speaker :-)
> 
> There's still a lot of duplication with the options page:
> 
> * a fair bit of text still appears to have been copied verbatim from the decks screen - are you planning to remove it, or what is your plan there?
> * listing out available options makes sense if you're planning to explain them in more detail, but I don't think it adds a lot of value if you're just listing the options, as the list can already be seen on the page

Thanks! Yes, it's still work in progress, it should be ready for review within the next few days (my time is really limited these days)

> 
> Also because this is turning into quite a few changes, it may be worth splitting it up into multiple pull requests, and focusing on one section at a time, instead of trying to tackle the entire chapter at once. 

Maybe that's complicated in this case, since the whole page is already mixed up, but I will definitely do in future PR if it's easier for you.

cqg2:
Ready for review.

dae:
Sorry, I missed your reply - will get back to you.

cqg2:
> Sorry, I missed your reply - will get back to you.

No worries. Thanks

dae:
I've made some tweaks and will merge this in now - thank you for your contributions!

PR #62, opened 2021-07-15
Fix broken links

RumovZ:
Shall I also make a commit to enable the linkcheck?


dae:
Previously the mdbook action was failing when linkcheck was enabled, but I just gave it a try, and it seems to work now. Thanks!

dae:
Unfortunately it did break things - the action succeeds, but the website goes offline. I'll need to run linkcheck in a separate CI job, which I can sort out once we get checking working properly.

RumovZ:
That's annoying. I hope you can make it work without it being more trouble than it's worth!

PR #63, opened 2021-07-19
Fix addon-docs links and anchors

RumovZ:
In theory, all links should be valid with this. 🙂

### DIFF
--- a/src/contrib.md
+++ b/src/contrib.md
@@ -56,7 +56,7 @@ them about it.
 
 ## Sharing Add-ons
 
-Please see <https://addon-docs.ankiweb.net/#/sharing>
+Please see <https://addon-docs.ankiweb.net/sharing.html>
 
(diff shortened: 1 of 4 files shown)


dae:
Thanks Rumo! The link checking seems to work well:

https://github.com/ankitects/anki-manual/runs/3109644425?check_suite_focus=true

Maybe even too well :-) I had to restart the test a few times after merging this, due to flaky connections to hosts like microsoft.com. Will wait a while to see if it's a temporary issue with GitHub actions; we could try moving the check to Buildkite if problems persist.

https://github.com/ankitects/anki-manual/runs/3109665609?check_suite_focus=true

RumovZ:
Hm, strange. I haven't seen anything like this on my machine. Might it help to tweak the client configuration used for sending the HEAD and GET requests?
But I'm glad it works overall!

PR #66, opened 2021-07-29
corrections and English clarified

mattsmi:
A couple more markdown files corrected. 
There are some good definitions in deck-options.md now. But a Mature Card is not defined. Should it be?

Second question: Spelling is mixed in some of the markdown files. Are you standardising on British (which includes Australian) or US English? I have assumed the former, given some of the older content, which I assume was authored by your fine self.

Thanks again, Damien!

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -9,10 +9,10 @@ before changing them, as mistakes could reduce Anki's effectiveness.
 
 Deck options are accessed by:
 - Clicking the gear icon on the `Decks` screen.
-- Selecting a deck on the `Decks` screen, and then clicking `Options` 
+- Selecting a deck on the `Decks` screen, and then clicking `Options`
 at the bottom of the screen.
-- Clicking on `More` > `Options` while in review mode. 
-- Pressing `o` while in review mode. 
+- Clicking on `More` > `Options` while in review mode.
+- Pressing `o` while in review mode.
 
 This page describes the options shown in Anki 2.1.45+, when you have the v2 or
 v3 scheduler enabled. On older versions, some options will not be available, or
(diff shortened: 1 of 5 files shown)


src/deck-options.md
```diff
@@ -390,26 +390,26 @@ If you plug the resulting 65% into the interval modifier, you should
 find over time that your retention moves closer to your desired
 retention.
 
-One important thing to note however is that the tradeoff between time
+One important thing to note however is that the trade-off between time
 spent studying and retention is not linear: we can see here that to
-increase our retention by 5 percentage points, we’d have to study 35%
+increase our retention by 5 per cent, we would have to study 35%
```
dae, line 395: I'm not sure this change is correct?

https://thewritingbusiness.com/the-difference-between-percentages-and-percentage-points/

dae:
Thanks Matt! Australian is my mother tongue, but using American English when programming and writing translations seems to be more common, and American-English speakers seem to be in the majority, so I tend towards using that these days. I don't have strong preferences for the manual - we're all volunteers here, from different backgrounds, and documentation in a mix of dialects is better than none at all. :-)

dae:
Mature is documented here, but if you think it needs to be covered again, please add it by all means: https://docs.ankiweb.net/stats.html?highlight=mature#types-of-cards

src/deck-options.md
mattsmi: Yes, you are right! Sorry, I think the whiskey must have run out. :)
Will you revert this change, or should I do it and resubmit a pull request?
Many thanks for you time and sorry the error!

dae:
```suggestion
increase our retention by 5 percentage points, we would have to study 35%
```

The interface should give you an option to accept this suggested change, and then I can merge the PR.

dae: Nevermind, I can do that myself :-)

PR #71, opened 2021-09-20
Mention comma syntax for id searches

RumovZ:
https://forums.ankiweb.net/t/add-a-flag-without-changing-mod-field-or-other-way-to-create-automatically-large-filtered-decks/13510/


dae:
Thanks!

PR #84, opened 2021-11-16
Adding some links and examples (for frequently asked questions in the forums)

cqg2:


src/filtered-decks.md
```diff
@@ -190,7 +190,7 @@ In the v1 scheduler, the **custom steps** option allows you to override the
 home deck’s steps and provide your own steps instead. The provided steps apply to both
 cards being learnt, lapsed reviews, and reviews ahead of time. Please note, however, that
 from the [v2 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html)
-filtered decks no longer support custom steps.
+filtered decks [no longer support custom steps.](https://forums.ankiweb.net/t/filtered-deck-interval-problem/11834/8)
```
dae, line 193:
```suggestion
filtered decks [no longer support custom steps].(https://forums.ankiweb.net/t/filtered-deck-interval-problem/11834/8)
```

src/media.md
```diff
@@ -6,7 +6,12 @@ locations](files.md) section. When you add media within Anki, either by
 using the paperclip icon in the [editor](editing.md) or by pasting it into
 a field, Anki will copy it from its original location into the media
 folder. This makes it easy to back up your collection’s media or move it
-to another computer.
+to another computer. 
+
+When importing a media file, please, please try not to use filenames with 
+blank spaces. Instead of importing the file "my picture.jpg", rename it to
+"my_picture.jpg", otherwise, [problems may arise](https://forums.ankiweb.net/t/new-html-editor-problem-searching-media-files-with-blank-spaces/14857)
+when [searching](searching.md) for media files.
```
dae, line 14: I'm not sure there's a need for this, cqg. Anki should be able to handle the spaces fine, and most users never use Find+Replace on media files.  Asking the user to rename all their files with spaces just so that they can avoid some confusing behaviour in a rare case seems like a bit of an overreaction.

src/studying.md
```diff
@@ -157,14 +157,19 @@ estimates, you can disable them in Anki’s [preferences](preferences.md).
...
 from sticking together and always coming up for review on the same day. This fuzz
 will appear on the answer buttons when the [v3 scheduler](https://faqs.ankiweb.net/the-2021-scheduler.html) is enabled, so if
 you are using a previous version and you’re noticing a slight discrepancy between
 what you select and the intervals your cards actually get, this is probably the
 cause.
 
+Learning cards are also given up to 5 minutes of extra delay so that they 
+don’t always appear in the same order, but answer buttons won't reflect that. 
+
+It is not possible to turn this feature off.
```
dae, line 171: Recommend joining this to the previous paragraph.

dae (commented):
Thanks cqg, some comments below

src/studying.md
cqg2: Done

src/media.md
cqg2: Actually,  it seems to be bigger problem, not only related to the Find & Replace command: a standard search in 2.1.45+...


 ...won't find any file media with blank spaces  (2):


Doing the same with 2.1.44...


...works as expected. 



If the problem can't be fixed, an alert in the manual can be useful, otherwise the user can spend sometime trying to figure out what's the problem here (real story, haha). Maybe rewording the paragraph would work better, or do you prefer not to mention the issue at all in the manual / FAQ?

dae: Mentioning it is fine, but rather than begging the user to avoid spaces, I think we should say something more like this:

If your media filenames contain spaces or other special characters such as percentage signs, the way the filenames appear in the HTML editor will differ from the way the filenames appear on disk. For example, a file called `hello 100%.jpg` will appear as `hello%20100%25.jpg` in the HTML editor. Internally, Anki still uses the original filenames, so if you would like to modify the filename with Find&Replace, you would need to use the name as it appears on disk., not as it appears in the HTML editor. Exporting to a text file is another way to see the underlying representation.

cqg2: Yes, that's better. I've applied the suggested changes, adding "so if you would like to **search for the file or** modify the filename with Find&Replace" if that's ok. 

Thanks!

src/media.md
```diff
@@ -6,15 +6,25 @@ locations](files.md) section. When you add media within Anki, either by
...
-to another computer.
+to another computer. 
+
+If your media filenames contain spaces or other special characters such 
+as percentage signs, the way the filenames appear in the HTML editor will 
+differ from the way the filenames appear on disk. For example, a file called 
+`hello 100%.jpg` will appear as `hello%20100%25.jpg` in the HTML editor. 
+Internally, Anki still uses the original filenames, so if you would like to
+[search](searching.md) for the file or modify the filename with [Find&Replace](browsing.md#find-and-replace), you would 
+need to use the name as it appears on disk., not as it appears in the
```
dae, line 17:
```suggestion
need to use the name as it appears on disk, not as it appears in the 
```

src/media.md
```diff
@@ -6,15 +6,25 @@ locations](files.md) section. When you add media within Anki, either by
...
 folder. This makes it easy to back up your collection’s media or move it
-to another computer.
+to another computer. 
+
+If your media filenames contain spaces or other special characters such 
+as percentage signs, the way the filenames appear in the HTML editor will 
+differ from the way the filenames appear on disk. For example, a file called 
+`hello 100%.jpg` will appear as `hello%20100%25.jpg` in the HTML editor. 
+Internally, Anki still uses the original filenames, so if you would like to
+[search](searching.md) for the file or modify the filename with [Find&Replace](browsing.md#find-and-replace), you would
```
dae, line 16:
```suggestion
[search](searching.md) for the file or modify the filename with [Find&Replace](browsing.md#find-and-replace), you will 
```

dae (commented):
Thanks cqg! I'll fix a few mistakes in the text I provided, then merge this in.

PR #85, opened 2021-11-20
Replaced FAQ in Table of contents with direct link

SuperNici:
[Relevant forum post](https://forums.ankiweb.net/t/add-external-links-to-summary-md-manual/15075)

[Relevant question asked on the mdBook github](https://github.com/rust-lang/mdBook/issues/1354#issuecomment-973678843)

Big thanks to @rsapkf, I would not have figured this out without their help.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -49,5 +49,4 @@
 - [Miscellaneous](misc.md)
 
 [Contributing](contrib.md)
-[FAQs](faqs.md)
 [Resources](resources.md)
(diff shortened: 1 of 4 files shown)


dae:
I'm afraid I'm not really fond of this change, as it means we now need to keep the template in sync with what mdbook expects when we update to future mdbook versions in the future. If you find yourself accessing the faqs frequently, you can just type the start of it into your browser and chances are it will remember your previous visit.

SuperNici:
Haha, fair enough. I honestly figured as much after way too much time reading mdbook docs.

This is a pain to maintain...

PR #86, opened 2021-11-23
Adapt to newest Anki version

jorislimonier:
Adapt Linux installation instructions for version 2.1.49


dae:
Thanks for taking the time to contribute! This is something I'd prefer not to have to update each time a new release is made, so I've changed it to a generic XX instead.

jorislimonier:
Ok, thank you

PR #88, opened 2021-11-24
Document startup problem with multiple copies

metavee:
Adding a startup issue I encountered after updating Anki, and its solution.


dae:
Thanks!

PR #90, opened 2021-11-28
Add a section about text direction

abdnh:
This adds a section explaining how to change text direction in card templates (common question among users using RTL languages).


dae:
We do have a brief mention of it on https://docs.ankiweb.net/templates/fields.html#right-to-left-text already, but no harm covering it in styling too - thanks!

abdnh:
Oh, I missed that section.

PR #93, opened 2021-12-18
Update contrib with a trick to retrieve the initial name of a shared deck

SamuelAllain:
I thought this trick could help users and reduce support burden.


dae:
Thanks!

PR #94, opened 2021-12-19
Remove outdated sentence "[SuperMemo] only runs on Windows computers."

olavfosse:
SuperMemo also runs on Android, iPadOS and macOS, see https://play.google.com/store/apps/details?id=com.supermemo and https://apps.apple.com/us/app/supermemo-effective-learning/id982498980.


dae:
It's a bit more nuanced than that - the powerful computer version only works on Windows, and other platforms only get their simplified app that focuses on premium subscriptions.

olavfosse:
Understood, thanks for explaining!

PR #99, opened 2022-01-27
Fix browsing

RumovZ:
This solves some issues I discovered while translating browsing.md.


src/browsing.md
```diff
@@ -220,9 +223,9 @@ they can be found in a submenu of the context menu.
 | Name | Action |
 |-|-|
 | Change Deck | Move currently selected cards to a different deck. |
-| Set Due Date | Turn cards into review cards, and make them due on a certain date. This can be useful for moving cards forward or back a few days when your study schedule is interrupted. Entering a range like `60-90` will give cards an interval between 60 and 90 days, which can be useful when you've imported a bunch of material you already know.
+| Set Due Date | Turn cards into review cards, and make them due on a certain date. This can be useful for moving cards forward or back a few days when your study schedule is interrupted. Entering a range like `60-90` will give new cards an interval between 60 and 90 days, which can be useful when you've imported a bunch of material you already know.
```
RumovZ, line 226: Not sure if I got this right, but I think only new cards receive the interval from Set Due Date, whereas an existing interval won't be modified. Is that correct?

src/browsing.md
```diff
@@ -220,9 +223,9 @@ they can be found in a submenu of the context menu.
...
+| Reposition | Change the order new cards will appear in. You can find out the existing positions by enabling the _due_ column, as described in the [table](#cardnote-table) section above. If you run the reposition command when multiple cards are selected, it will apply increasing numbers to each card in turn. By default the number increases by one for each card, but this can be adjusted by changing the "step" setting. The **Shift position of existing cards** option allows you to insert cards between currently existing ones, pushing the currently existing ones apart. For instance, if you have five cards and you want to move 3, 4, and 5 between 1 and 2, selecting this setting would cause the cards to end up in the order 1, 3, 4, 5, 2. By contrast, if you turn this option off, 1 and 2 will get the same position number (and it will thus be unpredictable which of the cards with the same number comes up first). Please note that when enabled, any card with a higher position will be modified, and all of those changed cards will need to be sent the next time you sync. |
```
RumovZ, line 228: For your convenience, the change here was removing a space and a pipe at the end.

src/browsing.md
dae:
```suggestion
| Set Due Date | Turn cards into review cards, and make them due on a certain date. This can be useful for moving cards forward or back a few days when your study schedule is interrupted. Entering a range like `60-90` will make the selected cards due between 60 and 90 days from now. New cards will have their interval set to the same delay, but reviews will be rescheduled without changing their current interval, unless '!' is included at the end of the range.
```

dae: Yep, but it might be worth mentioning that reviews are still rescheduled

src/browsing.md
dae: Thanks :-)

src/browsing.md
```diff
@@ -81,6 +81,8 @@ to append the resulting search to the current search.
 If you regularly search for the same thing,
 you can save the current search by right-clicking the topmost item in the sidebar,
 choosing “Save Current Search” and typing in a name.
+You can also drag and drop any sidebar item onto this area to add an equivalent
+saved search, basically pinning it to the top.
```
dae, line 85:
```suggestion
saved search, effectively pinning it at the top.
```

This wording may just be personal preference

src/browsing.md
```diff
@@ -234,37 +237,38 @@ parts of the browser, and to go up and down the card list.
...
-- Right click on selected notes in the _card list_, and click **Find and Replace**.
+The first input field is for the text that is going to be replaced, the second
+one for the replacement. Next, there is a dropdown menu that allows you to specify
+where Anki should look for text to replace: in a note's tags (requires Anki 2.1.45+),
+in all fields, or just in a specific field (only fields belonging to a selected
+note will be listed).
 
-- In the **Browser** window, click **Notes** in the menu bar and click **Find and Replace**.
+By default, only selected notes will be affected. If you want to lift that
+restriction, you can untick the first checkbox (requires Anki 2.1.45+).
```
dae, line 250:
```suggestion
restriction, you can untick the "selected notes only" checkbox (requires Anki 2.1.45+).
```

Perhaps naming the checkbox instead of referring to its position is a bit less likely to fall out of sync when changes are made to the interface in the future?

dae (commented):
Rest all looks good 👍

RumovZ:
cc03a41 is exactly your suggestion, not sure how the trailer got lost. Makes sense to be a bit more detailed there. 👍

dae:
Thanks Rumo :-)

PR #101, opened 2022-02-14
Change "Set Due Date" operation name.

alexeygorelov:
Is the description of that operation correct? I can't exactly understand what it means: "Make cards review cards".


dae:
Thanks for catching that! I've tweaked the description a bit to hopefully make it a bit more clear.

PR #102, opened 2022-02-14
Add instructions for searching for &, >, and <

Twinov:
I noticed that there isn't a way to search for the `&`, `>`, and `<` characters in Anki. Quoting or using `re:` on them doesn't work, and using a backslash to escape them gives you an invalid search error. But I found that using the HTML entity names works to search for cards containing them, so it's probably useful to put in the manual for anyone else in the future who comes across the issue.


dae:
Thank you!

PR #105, opened 2022-03-10
Update searching.md

RumovZ:


src/searching.md
```diff
@@ -405,5 +409,5 @@ Note and card IDs can be found in the [card info](stats.md) dialog in the
 browser. These searches may also be helpful when doing add-on
 development or otherwise working closely with the database.
 
-Object IDs will not work in the mobile clients, and are not intended to
+Object IDs will not work in AnkiMobile, and are not intended to
```
RumovZ, line 412: Is this still true for AnkiMobile? AnkiDroid supports it.

src/searching.md
```diff
@@ -284,10 +292,6 @@ cards that have moved into relearning more than 3 times
 `prop:ease!=2.5`  
 cards easier or harder than default
 
-Note that due only matches review cards and learning cards with an
-interval of a day or more: cards in learning with small intervals like
-10 minutes are not included.
-
```
RumovZ, line 290: This must have changed at some point, but I didn't investigate with which version.

src/searching.md
dae: They've been supported for some time, we could probably just drop this paragraph.

dae:
Thanks Rumo!

PR #107, opened 2022-03-20
Update manual to state it only sync on full sync

Arthur-Milchior:
I used to press sync to do backup before trying some change that may break my collection. I just discovered that I broke my collection and have no fork of the last 4 hours of reviews.
This update would not have helped me as I would not have read it, but at least new reader have up to date information


dae:
Will come back to this after the backup changes in beta 9 have had some testing/feedback, since we'll want to mention them as well.

PR #108, opened 2022-03-25
Add chapter on template checks and errors

RumovZ:
https://github.com/ankitects/anki/issues/1732 (1/2)
As always, improvements welcome.


dae:
Thanks Rumo! I made a few tweaks to try and make the process a bit clearer, but there's still plenty of room for improvement here, and we can make further tweaks as necessary.

PR #111, opened 2022-04-17
Warn about format changes interpreted as word boundaries

roxgib:
See here for how the current behaviour when using `w:` can cause unexpected matches:
https://forums.ankiweb.net/t/incorrect-browser-search-results/19156/4


dae:
Thank you!

PR #113, opened 2022-04-21
Add example for field search with spaces

RumovZ:
Quotation in field searches seems to be a cause for confusion. Not the first question regarding this: https://forums.ankiweb.net/t/find-two-slash-appearances-in-a-field/19293

### DIFF
--- a/src/searching.md
+++ b/src/searching.md
@@ -104,7 +104,7 @@ find notes with a Front field of exactly "dog". A field that says "a
 dog" will not match.
 
 `"animal front:a dog"`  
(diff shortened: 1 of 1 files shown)


src/searching.md
```diff
@@ -103,6 +103,10 @@ match' by default.
 find notes with a Front field of exactly "dog". A field that says "a
 dog" will not match.
 
+`"animal front:a dog"`  
+find notes with an Animal Front field of exactly "a dog". The double quotes are
```
dae, line 107:
```suggestion
find notes where the the "Animal Front" field is exactly "a dog". The double quotes are
```

PR #116, opened 2022-08-12
Fix conditional tts for ankidroid

libinvarghese:
Selector `:not(.android) .ankidroidtts`, will hide the content even if it is ankidroid, because of the way `:not()` works. Refer [JSFiddle](https://jsfiddle.net/0faxheou/)

It is best to specify the parent tag (`html`) on which class `android` is specified.

### DIFF
--- a/src/templates/fields.md
+++ b/src/templates/fields.md
@@ -137,7 +137,7 @@ Then in the styling section:
 .android .ankitts {
   display: none;
 }
-:not(.android) .ankidroidtts {
+html:not(.android) .ankidroidtts {
   display: none;
 }
 ```


dae:
Thank you!

sudomain:
Sorry if this is the wrong place for this, but with this workaround used in a Basic model, I have the text of my Front field displaying on AnkiDroud, but no text displayed on desktop (only the audio replay button):

Screenshot_20221228-141111

TTS is working correctly on both. I'd like to have the text of the Front field displayed on both platforms. Should I (and other users) be hiding the `.ankidrodtts` selector entirely using:
```
.ankidroidtts {
  display: none;
}
```

and then putting an additional {{Front}} field in our templates? It seems to allow the user to see the text of the field, but I haven't been able test if it interferes with the TTS on either platform. If this is a solution, should I submit a PR?

dae:
{{tts:Front}} never shows text, so you're supposed to add {{Front}} as well if you want to see it as well as hear it. Maybe the original poster wanted to hear but not see the text, leading to the confusion here? If the docs can be improved, PRs are always welcome.

PR #118, opened 2022-08-29
math.md: Typo and minor fixes

gnnoh:
I think there's an issue [here](https://github.com/ankitects/anki-manual/blob/bd853086b19dd7cf1a32018d7ee7912cd9737d58/src/math.md?plain=1#L74). If "Install missing packages on the fly" is set to "No", MikTex will not automatically fetch and install missing packages when Anki wants to render a document with new packages, causing a fatal error raised by MikTex. If it is set to "Always", everything works well. The problem is also mentioned in the [LaTeX Stack Exchange Community](https://tex.stackexchange.com/a/471185).

### DIFF
--- a/src/math.md
+++ b/src/math.md
@@ -25,7 +25,7 @@ To try it out:
 
         \(\sqrt{x}\)
 
-4.  If you click the Cards…​ button, you’ll see a preview of how the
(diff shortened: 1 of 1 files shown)


dae:
Thank you! I'm guessing that part of the docs was written from (incorrect) memory.

PR #121, opened 2022-09-03
Fix mistakes

ivan-pan:
I corrected a few mistakes such as `see the the` (instead of `see the`).

**Note:** The right arrow on the page at https://docs.ankiweb.net/platform/windows/installing.html leads to the very same page rather than the next one.


dae:
Thanks for these fixes Ivan!

PR #123, opened 2022-11-13
Add argument to use unzstd to uncompress

lifetheater57:
Add argument to use unzstd to uncompress the installation folder. Without `--use-compress-program=unzstd` it would fail on my machine.


dae:
Thanks Étienne, I've added a note about this.

PR #124, opened 2022-12-26
Added {{CardFlag}} to fields.md

Aquafina-water-bottle:
It seems like `{{CardFlag}}` was missing from the documentation, so I quickly added it in.

P.S. Would you be against adding the `{{furigana:...}` / `{{kana:...}}` / `{{kanji:...}}` to the documentation? I remember initially thinking that these required an addon to work, but it turns out it's supported natively by Anki. If not, I can make another PR to add it in.

### DIFF
--- a/src/templates/fields.md
+++ b/src/templates/fields.md
@@ -154,6 +154,8 @@ There are some special fields you can include in your templates:
 
     The card's subdeck: {{Subdeck}}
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Austin! No objections to adding documentation for those filters if you're motivated to do so.

PR #127, opened 2022-12-31
document TTS platform discrepancy and fix

sudomain:
AnkiDroid renders {{tts FieldName}} as text while other platforms render a (re)play TTS button. We should hide this text if we're on Android. To give AnkiDroid users a (re)play TTS button, we can make one that calls the JS API, and hide this button for all other platforms.

Ironically, I had to copy and paste this on mobile, so there may be mistakes.

Edit: a small thing to note is that this code renders text in the {{Front}} field on the platforms I tested (AnkiDroid and desktop (Linux). The occurency of {{Front}} which is responsible for this is the one surrounded by AnkiDroid's `<tts>` tags. If this text is undesirable, it also be hidden using CSS. I just wasn't sure if I should include that as well since it would result in front template that was visually blank.


dae:
LGTM - thanks @sudomain!

PR #130, opened 2023-02-03
Discard erroneous backslashes

snoyes:
Didn't realize in the previous commit that backslashes are not required inside a code block


dae:
Thanks for the quick fix!

PR #132, opened 2023-03-02
delete duplicated text, reorganize paragraphs, adding Tools>Check media options, adding some images

cqg2:
Some minor corrections, adddings,  and text reorganizing. I've also started to add some screenshots in editing.md, to try to make the manual more visual and attractive. Many more can be added all over the manual, but I'm not sure if you find this a good idea, so please let me know what you think about that.


dae:
Thank you!

Screenshots have a high maintenance cost, so I'm ok with adding them when it makes things considerably clearer for users, but I don't think we should be trying to add screenshots of every feature.

cqg2:
Yes, I absolutely agree, the idea is to add pictures only when it can be helpful. Thanks!

PR #136, opened 2023-04-17
Language changes to anki-manual/src/backgrouund.md

Iqbalmde:
@dae hey, did some minor changes to how the manual presented the definition of active recall and spaced repetition , if you find the changes meaningful, do accept my pull request, Thanks

### DIFF
--- a/src/background.md
+++ b/src/background.md
@@ -31,12 +31,12 @@ Understanding how they work will make you a more effective learner.
 
 ## Active Recall Testing
 
-_Active recall testing_ means being asked a question and trying to
-remember the answer. This is in contrast to _passive study_, where we
-read, watch, or listen to something without pausing to consider if we
-know the answer. Research has shown that active recall testing is far
-more effective at building strong memories than passive study. There are
-two reasons for this:
+_Active recall testing_ is a powerful study technique that involves retrieving 
+information from memory through back and forth being asked questions and providing answers. 
+This is in contrast to _passive study_, where we read, watch, or listen to something 
+without pausing to consider if we know the answer. Research has shown that active recall
+testing is far more effective at building strong memories than passive study. 
+There are two reasons for this:
 
 - The act of recalling something _strengthens_ the memory, increasing
     the chances we’ll be able to remember it again.
--- a/src/background.md
+++ b/src/background.md
@@ -83,6 +83,11 @@ if you only have 30 of them to review, but as the number grows to 300 or
 
 ## Spaced Repetition
 
+The idea behind _spaced repetition_ is that the brain needs time to process and consolidate 
+information, but it also needs regular reminders to keep the information fresh in memory. 
+By spacing out reviews, the brain is forced to work harder to recall the information, 
+strengthening the neural connections that facilitate memory.
+
 The _spacing effect_ was reported by a German psychologist Hermann Ebbinghaus in 1885. He
 observed that we tend to remember things more effectively, if we spread
 reviews out over time, instead of studying multiple times in one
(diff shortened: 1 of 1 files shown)


dae:
Please bear in mind that I'm rather busy, and reviewing changes takes time (especially when you re-wrap a paragraph, causing the entire paragraph to show as changed in the diff). For this reason, it would be great if you could focus on changes that will bring significant value to people reading the manual. Minor wording tweaks may be a small improvement, but it's not the most efficient use of our time.

src/background.md
```diff
@@ -31,12 +31,12 @@ Understanding how they work will make you a more effective learner.
...
 ## Active Recall Testing
 
-_Active recall testing_ means being asked a question and trying to
-remember the answer. This is in contrast to _passive study_, where we
-read, watch, or listen to something without pausing to consider if we
-know the answer. Research has shown that active recall testing is far
-more effective at building strong memories than passive study. There are
-two reasons for this:
+_Active recall testing_ is a powerful study technique that involves retrieving 
+information from memory through back and forth being asked questions and providing answers.
```
dae, line 35: "back and forth being asked" sounds unnatural to me

src/background.md
```diff
@@ -83,6 +83,11 @@ if you only have 30 of them to review, but as the number grows to 300 or
 
 ## Spaced Repetition
 
+The idea behind _spaced repetition_ is that the brain needs time to process and consolidate
```
dae, line 86: The original goal of this section was to introduce the history of spaced repetition. While this paragraph makes sense in isolation, it feels a bit out of place chronologically.

src/background.md
Iqbalmde: makes sense i've changed it to reciprocating which fits better probably

src/background.md
Iqbalmde: the reason i considered adding a paragraph of description about spaced repetition was because there was one for active recall above and here just the history was mentioned, nonetheless i removed it as you felt it was out of place , i'm new to open source so learning through still and thought it would be good to start contributing to learn , sorry i was a nuisance ;-; , will try making better valued contributions from next time

src/background.md
```diff
@@ -31,12 +31,12 @@ Understanding how they work will make you a more effective learner.
...
 ## Active Recall Testing
 
-_Active recall testing_ means being asked a question and trying to
-remember the answer. This is in contrast to _passive study_, where we
-read, watch, or listen to something without pausing to consider if we
-know the answer. Research has shown that active recall testing is far
-more effective at building strong memories than passive study. There are
-two reasons for this:
+_Active recall testing_ is a powerful study technique that involves retrieving 
+information from memory through reciprocating questions being asked and providing answers.
```
dae, line 35: I'm afraid I feel the original sentence was clearer than the replacement.

src/background.md
```diff
@@ -91,13 +91,11 @@ utilizing the spacing effect to improve learning, in what has come to be
...
-cards to a different box on each successful or unsuccessful review, it
-was possible to see at a glance a rough estimate of how well a card was
-known and when it should be reviewed again. This was a great improvement
-over a single box of cards, and it has been widely adopted by
-computerized flashcard software. It is a rather rough approach however,
+popularized a method of spaced repetition with paper flashcards. The system 
+involves organizing flashcards into boxes based on how well the user knows 
+the information, with more challenging cards being reviewed more frequently.
+This was a great improvement over a single box of cards, and it has been widely 
+adopted by computerized flashcard software. It is a rather rough approach however,
```
dae, line 98: I'm fine with the new wording in this paragraph if you feel it's worth changing.

Iqbalmde: It's okay , im working on some better improvements of manual which might take a while , ig we can for now close these commits , thanks for active replies, I'll update soon some better valued changes :)

PR #137, opened 2023-05-10
Update sync-server.md

sbadillo:
updated command for windows


dae:
The idea behind using anki-console was that the user would want to see a log of what was going on, but .exe was a mistake. Does the command work if you change it to .bat?

dae:
Ah, that's a shame; I'll fix the .bat file in the next update so that works. In the mean time, let's merge your PR so we show something that works to existing users. Thank you!

PR #138, opened 2023-07-26
Windows display issues: add PowerShell alternative for the Qt6 fix

YakovL:
This is a follow up to https://github.com/ankitects/anki/issues/1105#issuecomment-1598628783


dae:
Thanks!

PR #140, opened 2023-08-15
Add sentence about {{type:Field}} on ankiweb.net

ValentinSchwind:
The commit adds a sentence which mentions that type boxes will not be displayed on ankiweb.net.

I didn't add a description about how one might work around this (for e.g. by creating filtered decks, which collect all the typ-in-the-answer cards), because the section is already pretty long. I could however add this if it seems useful.


dae:
Thanks Valentin! I don't particularly see the need to cover workarounds there.

PR #141, opened 2023-08-15
Fixed a typo in background.md

MiVHQW:
practising --> practicing


MiVHQW:
So today I learned something new :D Then I would suggest to just keep it this way, since this is the UK version, which is more like the "official" standard.

PR #142, opened 2023-08-15
fixed typos: eg -> e.g.

MiVHQW:
eg to e.g.

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -95,7 +95,7 @@ Controls how many new cards are introduced each day you open the program. If you
 study fewer than the limit, or miss a day, the next day the counts will be back
 to your limit - they do not accumulate.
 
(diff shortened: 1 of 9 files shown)


dae:
I'm fine with this change, but there are a number of other cases in the manual where 'eg' has been used. So we don't have to create+review a whole bunch of separate PRs, it would be great if you could address the other usages of 'eg' in this same PR. Also please note you can push updates to this PR; you don't need to delete and create a new one.

MiVHQW:
I'd love to do that. I will update my PR with all the eg usages soon.

MiVHQW:
I've looked for all occurrences of eg and changed them to e.g. now ;)

dae:
Thanks Gabriel!

PR #143, opened 2023-08-18
Removed a leftover square bracket in preferences.md

RngValue:
`"Disable this if you don't want an automatic sync with Ankiweb when opening / closing a [profile](./profiles.md)]."`


dae:
Thanks!

PR #145, opened 2023-08-23
fixed interpunctuation

MiVHQW:


dae:
I'm no grammar expert, but I'm afraid I feel the originals are easier to read. Is this a necessary change?

MiVHQW:
According to my findings these commas are not correct by the usual standards
https://www.grammarly.com/blog/comma-before-and/

But of course the easy to read feeling is also a nice thing, so I will close this PR, since you can better feel this as a native speaker than me. :) I can just act according to the rules I know but sometimes they do not perfectly reflect the spoken/written language.

PR #149, opened 2023-09-03
Update documentation about interval after pressing Hard

user1823:
Reference: https://forums.ankiweb.net/t/interval-for-hard-is-longer-than-intervals-for-good-and-easy/24860/7

### DIFF
--- a/src/studying.md
+++ b/src/studying.md
@@ -73,8 +73,9 @@ There are four rating buttons when learning:
 - **Again** moves the card back to the first step.
 
 - **Hard** repeats the current step.
-  - If one or more steps are remaining, the delay will be the average of Again and Good, i.e., the average of the current step and the next step.
-  - If no steps are remaining, the delay is 50% larger than the current step.
+  - If the card is on the first (and the only) step, the delay is 50% larger than the step. But, this delay is at most one day larger than the step.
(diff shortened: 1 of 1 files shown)


dae:
That's not quite correct - what matters is if the user has 1 or more than 1 steps configured. Here's the relevant code:

    /// Special case the hard interval for the first step to avoid equality with
    /// the again interval. Also ensure it's smaller than the good interval,
    /// at least with reasonable settings.
    fn hard_delay_secs_for_first_step(self, again_secs: u32) -> u32 {
        if let Some(next) = self.secs_at_index(1) {
            // average of first (again) and second (good) steps
            maybe_round_in_days(again_secs.saturating_add(next) / 2)
        } else {
            // 50% more than the again secs, but at most one day more
            // otherwise, a learning step of 3 days and a graduating interval of 4 days e.g.
            // would lead to the hard interval being larger than the good interval
            let secs = (again_secs.saturating_mul(3) / 2).min(again_secs.saturating_add(DAY));
            maybe_round_in_days(secs)
        }
    }

user1823:
> what matters is if the user has 1 or more than 1 steps configured. Here's the relevant code:

Thanks for pointing me to the relevant code. I have made the appropriate correction.

dae:
Thanks!

PR #150, opened 2023-09-05
Add a Dependabot config to keep GitHub action versions updated

kurtmckee:
The `actions/checkout` action was recently updated to version 4.

Rather than submit a PR to update it once, this PR introduces a Dependabot config that will keep all GitHub actions updated on a regular basis.

If this PR merges, you can expect Dependabot to immediately submit at least one PR to update `actions/checkout@v3` to `v4`.


dae:
Sounds good - thanks!

PR #154, opened 2023-09-15
Update section about Qt6 video drivers

abdnh:
Related:
https://github.com/ankitects/anki/pull/2643
https://github.com/ankitects/anki/pull/2660


dae:
Thanks!

PR #157, opened 2023-09-24
Add details about prop:cdn and prop:cds

user1823:
Fixes https://forums.ankiweb.net/t/anki-manual-doesnt-talk-about-searching-in-custom-data/33707

Because FSRS is no longer going to depend on custom data, feel free to update the examples to suit the newer use cases.


dae:
Thank you!

PR #158, opened 2023-10-14
Document [anki:tts lang=en_US] tts tags

sudomain:
Introduced in 2.1.50. I'm unsure of Anki mobile's support of it, but AnkiDroid 2.16.5 seems to display a replay button for it but won't actually play it.

### DIFF
--- a/src/templates/fields.md
+++ b/src/templates/fields.md
@@ -65,7 +65,7 @@ each line, you would use
     {{Field 1}}<br>
     {{Field 2}}
 
-## Text to Speech
+## Text to Speech for individual fields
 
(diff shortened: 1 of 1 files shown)


src/templates/fields.md
```diff
@@ -158,6 +158,19 @@ html:not(.android) .ankidroidTtsButton {
 }
 ```
 
+## Text to Speech for multiple fields and static text
+
+This feature requires Anki 2.1.50. AnkiDroid does
```
dae, line 163:
```suggestion
This feature requires Anki 2.1.50+ or AnkiMobile 2.0.84+ . AnkiDroid does
```

dae:
Thanks for doing that!

PR #160, opened 2023-10-17
Add and reword link to installation issues for Windows.

thomaslantern:
Added a link to installation issues page for either mobile users (who would not see the links to the left, if at all), or users who have (intentionally or otherwise) closed their table of contents.

PR #161, opened 2023-10-18
Add and reword links to installation/startup issues

thomaslantern:
For reasons similar to https://github.com/ankitects/anki-manual/pull/160, added a link to installation issues and startup issues page for either mobile users (who would not see the links to the left, if at all), or users who have (intentionally or otherwise) closed their table of contents.

### DIFF
--- a/src/platform/windows/installing.md
+++ b/src/platform/windows/installing.md
@@ -65,4 +65,6 @@ you can download older Anki versions from the [releases page](https://github.com
 ## Problems
 
 If you encounter any issues when installing or starting Anki, please see the
-following links on the left.
+following pages:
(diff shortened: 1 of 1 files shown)


thomaslantern:
(I'm not sure this is the place to put this, but it seemed relevant and I couldn't figure out how to "pull request" this one.)
The link on the right button at the bottom of https://docs.ankiweb.net/platform/windows/installing.html is incorrect.
It should be https://docs.ankiweb.net/platform/windows/installation-issues.html, but it links back to https://docs.ankiweb.net/platform/windows/installing.html (i.e. the same page).

If there's a way for me to address this in the future and I just missed it, I would be happy to learn how (to save you the trouble!)

dae:
That is a bug in mdbook, and happens because we don't have separate pages for the mid-level items like "Windows".

PR #170, opened 2023-10-30
Minor grammar fix

GGGariKKK:
Added lacking word to complete passive infinitive form


GGGariKKK:
Have a question. If there is a number of corrections across multiple markdown pages, is it better to create PRs with corrections only for a single page, or it does not matter?

dae:
Thank you! And either would be fine.

PR #172, opened 2023-11-02
fix hyperlink to edit page

thomaslantern:
fixed hyperlink to take user to proper editing page (rather than just the "studying" main page)

### DIFF
--- a/src/editing.md
+++ b/src/editing.md
@@ -269,7 +269,7 @@ marks or fields are better.
 
 ## Editing Features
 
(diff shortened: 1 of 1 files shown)


dae:
Feel free to do a number of these in the same PR if that's easier.

thomaslantern:
@dae - thank you. It's usually easier doing one at a time, but I will definitely keep that in mind if it comes up - I appreciate it!

PR #179, opened 2023-11-09
Minor semantic correction

GGGariKKK:
I assume 'note type' was meant: 'Cards from the current note type' can be placed in separately specified decks with the help of 'Deck Override' option


dae:
Sorry, this is not correct. 'card type' is a synonym for 'card template'. Deck override can't select a different notetype to the one that is active.

GGGariKKK:
No problem, thanks for feedback

PR #180, opened 2023-11-10
Update info about desired retention and the chart.md

Expertium:


dae:
I've manually made this change, as your image link was broken, and the old image was not removed. I didn't include the line about the limits, as the UI already enforces them, so I think mentioning them there is not contributing much.

PR #192, opened 2024-01-06
Improve sentence clarity and fix typos

hnvy:
Hello,

First, happy new year!

I went over the [new additions](https://github.com/ankitects/anki-manual/compare/1db23cd..f15752f#diff-13dce136e586859a8a3aae45c092bfc1588848d145719f6f8f37931bc5000131) since the 1db23cd70d61c20096bb74ae240e4d063c344a1d.

Not much has come up other than this small typo and sentence.

I have added the markdown grave accents around the buttons.

There are few more questions, if you do not mind:
1. "[Log-loss](https://github.com/ankitects/anki-manual/blob/main/src/deck-options.md?plain=1#L539)" in the official manual: what is this referring to exactly? Review log? Or [_Log_ arithm](https://en.wikipedia.org/wiki/Logarithm)? I have translated it as "Review log", but now I am thinking that it might be referring to _Log_ arithm.
2. This question is related to the [following line](https://github.com/ankitects/anki/blob/3982e0c8fe691d3e84deceb7555f2d2fd0a8fbe6/ftl/core/preferences.ftl#L26) in Anki. This is the only British "Synchroni **s** ation" in Anki. All of the others are American. I wanted to open a PR, but I thought it may not be worth it (hence why I asked first).

As always, many thank for all of your hard work last year.

HR

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -392,7 +392,7 @@ The options are:
 Requires Anki 23.12 or later. Auto Advance allows you to automatically reveal
 the answer and/or move to the next card. To use it, you must first set a non-zero
 time in "seconds to show question" and/or "seconds to show answer". Then, in the
-review screen, use the Auto Advance action from the More button to start advancing.
+review screen, use the Auto Advance action from the `More` button to start advancing.
 
 ## Burying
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Harvey!

- Log loss is a technical term: https://www.analyticsvidhya.com/blog/2020/11/binary-cross-entropy-aka-log-loss-the-cost-function-used-in-logistic-regression/
- Happy to accept PRs that change Britishisms, as they should be in the en-GB template instead. They snuck in because I'm Australian, and was not careful about this in the past.

PR #195, opened 2024-01-23
Add Add-ons page to manual

NameLessGO:
Wasn't sure in how to implement Tools -> add-ons in the same paragraph, so had to remove it

https://forums.ankiweb.net/t/add-ankiweb-feature-to-browse-or-search-for-add-ons/40145/4

### DIFF
--- a/src/addons.md
+++ b/src/addons.md
@@ -6,6 +6,8 @@ scheduling, and so on.
 
 To browse the list of available add-ons, access https://ankiweb.net/shared/addons
 
+Alternatively, select  ``Tools`` -> ``Add-ons``, then click on ``Get Add-ons``
+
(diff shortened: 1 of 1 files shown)


dae:
Why remove the previous instructions, instead of add to them? Surely it's easier for users to open the add-ons screen than it is to search for the link in the manual.

NameLessGO:
Sorry, after brainstorming the idea still couldn't find a satisfied way to implement so added another paragraph

user1823:
Suggestion:

To browse the list of available add-ons, select `Tools` → `Add-ons`, then click on `Get Add-ons`. Alternatively, open https://ankiweb.net/shared/addons in a web browser.

PR #196, opened 2024-01-23
add libsmime3.so for chromebook install

NameLessGO:
Fixes ImportError: libsmime3.so: cannot open shared object file: No such file or directory" in Chromebook

https://forums.ankiweb.net/t/for-cb-users-i-suggest-reading-this-post-to-install-anki-linux-in-cb-instead/40149


dae:
Thank you!

PR #199, opened 2024-01-24
Changed version schemed in installation from 2.1XX to 23.XX

NameLessGO:
Another idea could be 2X.XX for future-proof, 24,25...

But thought it would be more confusing, and we can just change the value next year

### DIFF
--- a/src/platform/linux/installing.md
+++ b/src/platform/linux/installing.md
@@ -31,8 +31,8 @@ To install Anki:
 3. Open a terminal and run the following commands, replacing the filename as appropriate.
 
 ```shell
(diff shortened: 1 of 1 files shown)


dae:
I'd rather not have to change the manual every year, and users may select a release from a previous year when they don't upgrade immediately. How about just removing the version number completely?

tar xaf Downloads/anki-XXX-linux-qt6.tar.zst

NameLessGO:
Not a fan of XXX, 2XXX indicates to start the change filename after 2 (2.1XX for past versions and 2X.XX for newer)

dae:
That'll work

PR #200, opened 2024-02-03
Corrected menu item names

Danika-Dakika:
Throughout the manual "notetype/s" and "note type/s" are often used interchangeably. I'm not taking a stance on which is correct, so I'm not trying to edit them all -- but these at least should match the item name in the actual menu.


dae:
I'd like to switch to all-'noteype' in the future, but agree the items should match the menu entries. Thanks!

PR #201, opened 2024-02-09
removed random -- and bug with next page

NameLessGO:
Reported on Discord by mod khonkhortisan

Next chapter > links back to the same page for these pages: Installing & Upgrading Anki on [Windows](https://docs.ankiweb.net/platform/windows/installing.html) [macOS](https://docs.ankiweb.net/platform/mac/installing.html) [Linux](https://docs.ankiweb.net/platform/linux/installing.html)

[Preferences](https://docs.ankiweb.net/preferences.html) has a hr at the bottom

About the Next chapter issue, seems like Docs can't recognize 2.1.x, and it returns the main page, Windows 2.1

Runs perfectly here: https://doc.rust-lang.org/nightly/reference/expressions.html

Part of the code that handles Previous/Next (unsure on how to implement here)
https://github.com/rust-lang/mdBook/blob/5a4ac03c0de866c4cf30d41dd0c7ea02d8429a7a/guide/src/format/theme/index-hbs.md?plain=1#L80

### DIFF
--- a/src/preferences.md
+++ b/src/preferences.md
@@ -175,5 +175,3 @@ For info on the custom sync server option, see [this section](./sync-server.md).
 ## Backups
 
 Please see [this](backups.md#automatic-backups) section of the manual.
-
----


dae:
The next/previous chapter links don't work because mdbook doesn't handle things correctly when we skip the separate  subchapter pages.

NameLessGO:
@dae Don't think it is the case anymore since mdbooks runs fine

https://rust-lang.github.io/mdBook/format/theme/index-hbs.html?highlight=next#2-previous--next


Could we make an issue here to avoid duplicate in the future?

dae:
Yes, feel free to make an issue. The issue is limited to the platform notes section, which doesn't have separate entries for e.g. "Windows".

NameLessGO:
https://github.com/ankitects/anki/issues/3005

PR #205, opened 2024-03-09
Added missing links to Windows problem pages

thewilliamdrew:
The Windows Install and Upgrade page is missing some links to problem pages in the relevant subsection.


dae:
Thanks William! That section is about problems installing/starting Anki, so perhaps we don't want the text size and copy/paste links there?

thewilliamdrew:
Makes sense. Omitted.

dae:
Thanks!

PR #207, opened 2024-03-10
Importing without scheduling data removes leech and marked tags

arbyste:
Adding a note to the importing section to indicate that importing without scheduling data also removes "leech" and "marked" tags from the imported notes.


dae:
Thank you!

PR #209, opened 2024-03-22
Document ability to store passwords hashed

laalsaas:
As introduced by ankitects/anki#3083

### DIFF
--- a/src/sync-server.md
+++ b/src/sync-server.md
@@ -87,8 +87,8 @@ cargo install --path rslib/sync
 
 ## Multiple Users
 
-SYNC_USER1 declares the first user and password, and must be set.
-You can optionally declare SYNC_USER2, SYNC_USER3 and so on, if you
+`SYNC_USER1` declares the first user and password, and must be set.
+You can optionally declare `SYNC_USER2`, `SYNC_USER3` and so on, if you
 wish to set up multiple accounts.
 
 ## Storage Location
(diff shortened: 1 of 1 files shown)


src/sync-server.md
```diff
@@ -87,10 +87,20 @@ cargo install --path rslib/sync
...
 
-SYNC_USER1 declares the first user and password, and must be set.
-You can optionally declare SYNC_USER2, SYNC_USER3 and so on, if you
+`SYNC_USER1` declares the first user and password, and must be set.
+You can optionally declare `SYNC_USER2`, `SYNC_USER3` and so on, if you
 wish to set up multiple accounts.
 
+## Hashed Passwords
+
+If you want to store your passwords hashed, you can set the environment
```
dae, line 96: Some users may not know if they want that or not. How about the following wording:

Advanced users may wish to use hashed passwords instead of plain text passwords. If you wish to do this,
you'll need to use a separate tool (such as https://git.sr.ht/~laalsaas/pbkdf2-password-hash) to generate
a password hash. You can then tell the server to expect hashed passwords by setting the env var PASSWORDS_HASHED to 1.

When hashed passwords are used, `SYNC_USER` variables are expected to be in `username:password_hash` format, where password_hash is a hash of the password in the [PHC
Format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md).

laalsaas: I mostly adopted your wording, with some small changes.

dae:
Thanks!

PR #210, opened 2024-04-03
Update preferences.md to reflect v3 scheduler availability

AntoineQ1:
I updated the text to reflect the availability of v3 scheduler on all platforms.

_Don't hesitate to point out if this change needs to be phrased differently._

### DIFF
--- a/src/preferences.md
+++ b/src/preferences.md
@@ -90,7 +90,7 @@ Disable this option to hide the card count at the bottom of the screen.
 Useful to know how far in the future your cards are being pushed.
 
 **Spacebar (or enter) also answers card**\
-Defines whether a press on the space bar (or the enter key) also answers card.
+Defines whether a press on the space bar (or the enter key) also answers cards.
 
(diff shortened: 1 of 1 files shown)


dae:
Your change is fine, but given that we don't actually show that option in the preferences screen anymore, maybe we should just remove mention of it instead?

AntoineQ1:
Yes, absolutely, I didn't think about that !

src/preferences.md
```diff
@@ -106,6 +89,9 @@ Disable this option to hide the card count at the bottom of the screen.
 **Show next review time above answer buttons**\
 Useful to know how far in the future your cards are being pushed.
 
+**Spacebar (or enter) also answers card**\
+Defines whether a press on the space bar (or the enter key) also answers card.
```
dae, line 93:
```suggestion
Defines whether a press on the space bar (or the enter key) also answers cards.
```

dae:
Thank you!

PR #215, opened 2024-07-13
update studying.md

brishtibheja:
Fixed punctuation and articles, improved brevity of a sentence, added Oxford commas because it is used elsewhere in the document. Changed "give up" to "move on".

Dae, I noticed you use `""` at times and sometimes `''`. The US style is heavily used in the web but single inverted commas are easier to read and is recommended by Oxford's style guide. Is there a particular style you prefer while using inverted commas?

### DIFF
--- a/src/studying.md
+++ b/src/studying.md
@@ -10,20 +10,19 @@ to start studying.
 Study in Anki is limited to the currently selected deck as well as any
 subdecks it contains.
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks, changes look good. I think double quotes are what I'd typically reach for these days, and the type of the outer surrounding quotes is probably what influenced the selection of the inner quotes in the past. These days are strings are mainly in .ftl files, and double quotes in source code is less of a concern.

PR #219, opened 2024-07-27
Update searching.md

brishtibheja:


brishtibheja:
@dae It seems there was something called custom data the FSRS custom scheduling version used? Is that still needed? If not, searches for that should be removed. If other schedulers can still potentially use that, then people should be explained what is custom data. The manual seems to mention the phrase only thrice.

Danika-Dakika:
I have some suggested corrections and edits for this PR. I'm going to attempt to add them as comments on the changes themselves. We'll see if that makes it clearer than doing it from here.

src/searching.md
```diff
@@ -3,17 +3,16 @@
 <!-- toc -->
 
 Anki's Browse screen and the Filtered Deck feature use a common method
-of searching for specific cards/notes.
+of searching for specific cards/notes. This method is also used in the search bar below "FSRS parameters" field.
```
Danika-Dakika, line 6: At a minimum this needs an article -- below ***the*** "FSRS ... 

But I don't think it's quite fitting to call that a "search bar," because that suggests it returns a result to the user. I would suggest: 
`This method can also be used to adjust the scope of FSRS optimization.`

src/searching.md
```diff
@@ -3,17 +3,16 @@
...
 
 ## Simple searches
 
 When you type some text into the search box, Anki finds matching notes
 and displays their cards. Anki searches in all fields of the notes, but
-does not search for tags (see later in this section for how to search
-for tags). Some examples:
+does not search for tags (For information on how to search for tags, see [this section](##tags-decks-cards-and-notes) of the page). Some examples:
```
Danika-Dakika, line 12: Needs corrections to capitalization and/or punctuation (and the syntax of the link, I think). 

If the text is to be updated, in the manual, "section" is mostly used to refer to Summary/TOC items, and "page" usually refers to another site (like the FAQ). It's defnitely not 100%, but it would be better to stick with it and move toward consistency and coherency. 

But I don't think it is necessary to change the line at all to add this link -- just use the current text: 
`does not search for tags (see [later in this section](#tags-decks-cards-and-notes) for how to search for tags).`

src/searching.md
```diff
@@ -35,24 +34,24 @@ finds notes with neither "cat" nor "mouse".
 same as the above.
 
 `"a dog"`\
-finds notes with the exact sequence of characters "a dog" on them, such
+finds notes with the exact sequence of characters "a dog" in them, such
 as "atta dog", but not "dog a" or "adog".
 
 `-"a dog"`\
-finds notes without the exact phrase "a dog"
+finds notes without the following sequence of characters: "a dog".
```
Danika-Dakika, line 41: Since this is offered in contrast to the example above, better to mirror that line: 
`finds notes without the exact sequence of characters "a dog" in them.`

src/searching.md
```diff
@@ -35,24 +34,24 @@ finds notes with neither "cat" nor "mouse".
...
-finds notes with d, &lt;zero or more letters&gt;, g, like dg, dog, dung,
+finds notes with d, &lt;zero or more characters&gt;, g, like dg, dog, dung,
 etc.
 
 `w:dog`\
-search for "dog" on a word boundary - will match "dog", but not "doggy"
-or "underdog". Requires Anki 2.1.24+ or AnkiMobile 2.1.61+. Note that
-formatting changes may be interpreted as word boundaries e.g. searching
-for `w:exam` will match **exam**ple.
+searches for the word "dog" as opposed to a sequence of characters - will match "dog", but not "doggy"
```
Danika-Dakika, line 51: (Comment about "searches" above.)

If "word boundary" needs to be defined, it seems better to do that without removing the phrase itself, because it is somewhat of a term of art in search syntax. 
`search for "dog" on a word boundary (as a whole word, not a sequence of characters in another word) - will match`

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
 `front:dog`\
-find notes with a Front field of exactly "dog". A field that says "a
+finds notes with a Front field of exactly "dog". A field that says "a
 dog" will not match.
 
 `"animal front:a dog"`\
-find notes where the "Animal Front" field is exactly "a dog". The double quotes are
-mandatory, see [below](#matching-special-characters).
+finds notes where the "Animal Front" field is exactly "a dog". The double quotes are
+mandatory, see [this section](#matching-special-characters) of the page.
```
Danika-Dakika, line 106: This line is fine as-is. If it needs clarification, as mentioned above, it should be consistent with uses of "section" and "page" in the manual.

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
-find notes with the tag "animal", or subtags like "animal::mammal"
+finds notes with the tag "animal", or subtags like "animal::mammal".
 
 `tag:none`\
-find notes with no tags
+finds notes with no tags.
 
 `tag:ani*`\
-find notes with tags starting with ani
+finds notes with tags starting with ani.
```
Danika-Dakika, line 132: Should add quotation marks as well, to match the other examples:
`finds notes with tags starting with "ani".`

brishtibheja:
```suggestion
finds notes with tags starting with "ani".
```
You can add the suggestions like this.

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
 
 `deck:french`\
-find cards in a French deck, or subdecks like French::Vocab
+find cards in a deck called "French", or its subdecks like French::Vocab.
```
Danika-Dakika, line 135: Should add quotation marks to subdeck name as well, to match the other examples:
`or its subdecks like "French::Vocab".`

brishtibheja:
```suggestion
find cards in a deck called "French", or its subdecks like "French::Vocab".
```

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
-find cards in French, but not subdecks
+finds cards in French, but not subdecks.
 
 `deck:"french vocab"`\
-searching when a deck has a space
+searching when the deck name has a space.
 
 `"deck:french vocab"`\
-also ok
+same as earlier.
```
Danika-Dakika, line 144: Since it is about physical space "earlier" doesn't fit here. 
Alternative A:
`same as above.`
Alternative B: combine the examples with one label for both. 
```
`deck:"french vocab"`\
`"deck:french vocab"`\
searching when the deck name contains a space.
```

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
+filtered decks only.
 
 `-deck:filtered`\
-normal decks only
+normal decks only.
 
 `preset:"Default"`\
 cards in all decks that use the 'Default' deck options preset.
-Requires Anki 23.10+.
+Requires Anki(Mobile) 23.10+, or AnkiDroid 2.17+.
```
Danika-Dakika, line 154: I haven't seen anywhere else that Anki and AnkiMobile are collapsed into a single name like that. I don't think it makes it clear what we're talking about. 

The full list would be better:
`Requires Anki 23.10+, AnkiMobile 23.10+, or AnkiDroid 2.17+.`

I don't think it needs to be more succinct, but if nececessary, another option would be:
`Requires Anki or AnkiMobile 23.10+, or AnkiDroid 2.17+.`

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
-find notes with tags starting with ani
+finds notes with tags starting with ani.
 
 `deck:french`\
-find cards in a French deck, or subdecks like French::Vocab
+find cards in a deck called "French", or its subdecks like French::Vocab.
 
 `deck:french -deck:french::*`\
-find cards in French, but not subdecks
+finds cards in French, but not subdecks.
```
Danika-Dakika, line 138: Should add quotation marks as well, to match the other examples:
`finds cards in "French", but not its subdecks.`

src/searching.md
brishtibheja: Google uses "earlier". Referance: https://developers.google.com/style/word-list#above

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
+normal decks only.
 
 `preset:"Default"`\
 cards in all decks that use the 'Default' deck options preset.
-Requires Anki 23.10+.
+Requires Anki(Mobile) 23.10+, or AnkiDroid 2.17+.
 
 `card:forward`\
-search for Forward cards
+searches for cards with card type named "Forward".
```
Danika-Dakika, line 157: I think this one, along with 160 and 164, should be changed from "search/es for" to "finds" -- to match the other examples. 

It sounds a tiny bit off to say "cards with card type named "Forward"." That phrasing sounds like a card has multiple types, like we might talk about a note "with" certain card types, not like card type is an inherent property of the card.

There are different alternatives for this, but I can't necessarily explain why they sound better to me. 
`cards of card type "Forward".`
`cards that are card type "Forward".`
`cards with card type "Forward".` [Interestingly, maybe it's just the "named" that throws it off, because this sounds fine.]
`cards with the card type of "Forward".`

src/searching.md
brishtibheja: I think I missed updating some `Anki 23.10+` to `Anki(Mobile) 23.10+` but now I'm not sure whether that should be done. Everytime AnkiMobile is mentioned the version number is usually different from Anki. @dae how should this case be handled?

src/searching.md
brishtibheja:
```suggestion
finds cards in "French", but not its subdecks.
```

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
 
 `card:1`\
-search for cards by template number - eg, to find the second cloze
+searches for cards by card type number, e.g. to find the second cloze
 deletion for a note, you’d use `card:2`
 
 `note:basic`\
-search for cards with a Basic note type
+searches for cards with a note type named "Basic".
```
Danika-Dakika, line 164: This has the same issue as above -- it's just a little bit off, but for a slightly different reason.

Some other options:
cards with the note type "Basic".
cards created by note type "Basic".

src/searching.md
```diff
@@ -95,86 +94,86 @@ Things to note from the above:
...
 
 ## Ignoring accents/combining characters
 
 Requires Anki 2.1.24+ or AnkiMobile 2.0.60+.
```
Danika-Dakika, line 168: Is there an AnkiDroid version that could be added here to complete the page?

src/searching.md
```diff
@@ -183,148 +182,151 @@ Searches that ignore combining characters are slower than regular searches.
 Anki 2.1.24+, AnkiMobile 2.0.60+ and AnkiDroid 2.17+ support searching in notes with "regular expressions",
 a standard and powerful way of searching in text.
 
-Start a search with `re:` to search by regular expression. To make things easier, Anki will
+Start a search with `re:` to search using regular expression. To make things easier, Anki will
```
Danika-Dakika, line 185: If changing "by" to "using," the plural sounds more accurate. I can't explain why. 
`search using regular expressions.`

src/searching.md
```diff
@@ -183,148 +182,151 @@ Searches that ignore combining characters are slower than regular searches.
...
 `front:re:[a-c]1`\
-matches uppercase or lowercase a1, B1 or c1 that occurs anywhere in the "Front" field
+matches uppercase or lowercase a1, B1 or c1 that occurs anywhere in the "Front" field.
 
 `front:re:^[a-c]1$`\
-like the above, but will not match if any other text falls before or after a1/b1/c1.
+same as previous, but will not match if any other text falls before or after a1/b1/c1.
 
-Anki 2.1.50 added regex support for tags:
+Anki 2.1.50+ supports regular expressions for tags:
```
Danika-Dakika, line 205: Are there AnkiMobile and AnkiDroid versions that could be added here?

src/searching.md
```diff
@@ -183,148 +182,151 @@ Searches that ignore combining characters are slower than regular searches.
...
-cards in learning
+cards in learning.
 
 `is:review`\
-reviews (both due and not due) and lapsed cards
+reviews (both due and not due) and lapsed cards.
 
 `is:suspended`\
-cards that have been manually suspended
+cards that have been automatically or manually suspended.
```
Danika-Dakika, line 238: Can this line and the next (241) match? Either phrasing is fine.

src/searching.md
```diff
@@ -183,148 +182,151 @@ Searches that ignore combining characters are slower than regular searches.
...
 
 `is:buried`\
 cards that have been buried, either [automatically](studying.md#siblings-and-burying) or
-manually
+manually.
 
-Note that with the [new scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html),
+Note that with the [v2 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html),
```
Danika-Dakika, line 244: I think it's great to catch this sentence up with the times! Unfortunately, it now sort of sounds like that is only a feature of the (no longer supported) v2 scheduler, and doesn't necessarily carry forward to v3. 

`Note that as of the [v2 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html), Anki distinguishes`
`Note that starting with the [v2 scheduler](https://faqs.ankiweb.net/the-anki-2.1-scheduler.html), Anki distinguishes`

[On the other hand, this is a very strange place for this information -- a note about unburying in the middle of a section about search syntax. It might be better to move this to wherever unburying is documented.]

src/searching.md
```diff
@@ -183,148 +182,151 @@ Searches that ignore combining characters are slower than regular searches.
...
+cards that have moved into relearning more than 3 times.
 
 `prop:ease!=2.5`\
-cards easier or harder than default
+cards easier or harder than default.
 
-`prop:cdn:d>5` (Requires Anki 2.1.64+.)\
-cards with the value of `d` in custom data (usually refers to Difficulty in FSRS) greater than 5
+`prop:cdn:d>5`\
+cards with the value of `d` in custom data (usually refers to difficulty in FSRS) greater than 5 (requires Anki 2.1.64+).
```
Danika-Dakika, line 305: I think capitalizing "Difficulty" is correct here, since it's referring to one of the titled values of FSRS.

src/searching.md
```diff
@@ -335,47 +337,46 @@ included even if their notes were added a long time ago.
...
 
-This section was written for Anki 2.1.36+ - earlier versions did not support escaping
-characters in certain situations.
+If you're using a version earlier than Anki 2.1.36 the following searches may not work.
 
 As shown in the previous section, some characters like `*`, `_` and `"` have a
-special meaning in Anki. If you need to locate those characters in a search,
-you need to tell Anki not to treat them specially.
+special meaning in search. If you need to locate those characters in a search,
+you need to tell Anki not to treat them specially. This is called escaping a character and is primarily done by using double quotes and backslashes.
```
Danika-Dakika, line 374: `"escaping" `
or
`"escaping a character"`

src/searching.md
```diff
@@ -387,27 +388,26 @@ you need to tell Anki not to treat them specially.
   use `\\` instead of `\`.
 
 - `(` and `)`\
-  You can search for parentheses either by enclosing the full term in quotes,
-  and/or by using a backslash. That is, `"some(text)"`, `some\(text\)` and
-  `"some\(text\)"` are all equivalent, but `some(text)` is not.
+  You can search for parentheses by enclosing the entire term in quotes,
+   by using a backslash, or both at the same time. For example, `"(text)"`, `\(text\)` and
+  `"\(text\)"` are all equivalent searches, and searches for `(text)`.
```
Danika-Dakika, line 393: At the end, this should be:
`and search for `(text)`.`

But I don't agree with removing the more detailed examples that include "some" (unless they were incorrect in some way I'm not seeing?). I think they illustrate well that the entire term needs to be enclosed.

src/searching.md
```diff
@@ -387,27 +388,26 @@ you need to tell Anki not to treat them specially.
...
+  you can either use a backslash, or include the text in quotes. For example,
+  `\-free` or `"-free"` will match "guilt-free" and "cruelty-free".
 
 - `:`\
-  Colons have to be escaped unless they are preceded by another, unescaped colon.
-  So `w:e:b` is a word boundary search for `e:b`, `w\:e\:b` searches literally for
-  `w:e:b` and `w\:e:b` searches the field `w:e` for `b` (see
-  [field searches](#limiting-to-a-field)).
+  Colons have to be escaped using backslashes unless they are preceded by another, unescaped colon.
+  For example, `w:3:30` searches for 3:30 on word boundary and doesn't require you to use a backslash.
```
Danika-Dakika, line 403: Should use quotation marks, to match other examples:
`searches for "3:30" on word boundary`

brishtibheja:
```suggestion
  For example, `w:3:30` searches for "3:30" on word boundary and doesn't require you to use a backslash.
```

src/searching.md
brishtibheja:
```suggestion
  `"\(text\)"` are all equivalent searches, and search for `(text)`.
```
I believe the reader, if he has read other parts of this page/section will already understand what 'entire term' means. A simpler example is better here.

src/searching.md
brishtibheja:
```suggestion
you need to tell Anki not to treat them specially. This is called "escaping a character" and is primarily done by using double quotes and backslashes.
```

src/searching.md
brishtibheja: We don't capitalise DSR in the statistics screen of app. Here is the part of Sherlock's essay where he doesn't capitalise these terms: [Spaced Repetition Algorithm](https://github.com/open-spaced-repetition/fsrs4anki/wiki/Spaced-Repetition-Algorithm:-A-Three%E2%80%90Day-Journey-from-Novice-to-Expert#:~:text=Below%20are%20some%20empirical%20observations%20from%20Wozniak%2C%20corroborated%20by%20data%20from%20language%20learning%20platforms%20like%20MaiMemo%3A)

src/searching.md
brishtibheja: AnkiDroid 2.17 probably, I'm not sure about AnkiMobile. @dae?

src/searching.md
brishtibheja:
```suggestion
Start a search with `re:` to search using regular expressions. To make things easier, Anki will
```

I agree with you.

src/searching.md
brishtibheja: I was about to follow your advice here, but I remembered that some of the variability is intentional. It is a bit boring to read a text if everything seems to be using some sort of parallel syntax. I don't think this should be changed.

src/searching.md
brishtibheja:
```suggestion
Requires Anki 2.1.24+, AnkiMobile 2.0.60+ or AnkiDroid 2.17+.
```

src/searching.md
brishtibheja: I think if we don't include the word "named" people might think there are card types with an inherent property that makes them "Forward" card types. How about this:

```suggestion
searches for the cards of a card type named "Forward".
```

src/searching.md
brishtibheja:
```suggestion
searches for cards created with a note type named "Basic".
```

I'm using "with" because it is the user who will be creating these cards.

src/searching.md
brishtibheja: In line 53 we use the term "word boundaries" and later we slowly transition to more frequent use of this. I don't think we really need to define word boundaries here. It is bit hard to define, and I think translators will find this sentence easier to deal with than your suggestion.

src/searching.md
brishtibheja: I have mentioned below that some of this is intentional. This section/page of the manual has a lot of similar sounding sentences and it can quickly get boring to read through this. A little bit of variability, that catches the reader off guard, can draw their attention to what is written in the doc. I do value being consistent in formatting and else, but some amount of entropy is nice to have.

src/searching.md
```diff
@@ -3,17 +3,16 @@
...
 
 When you type some text into the search box, Anki finds matching notes
 and displays their cards. Anki searches in all fields of the notes, but
-does not search for tags (see later in this section for how to search
-for tags). Some examples:
+does not search for tags (For information on how to search for tags, see [this section](##tags-decks-cards-and-notes) of the page). Some examples:
 
 `dog`\
-search for "dog" - will match words like "doggy" and "underdog" too.
+searches for "dog" - will match words like "doggy" and "underdog" too.
```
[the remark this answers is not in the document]
brishtibheja: The conjugation will feel consistent otherwise. In the very next phrase we say "will match words" as if to say "the search term will match words like...". It is only appropriate to use searches instead of search, to mean "this term searches for 'dog'. Using  "search" will make it sound like an imperative sentence, as in, "Do this" or "Search that", or a TV ad "use our new...".

src/searching.md
brishtibheja:
```suggestion
does not search for tags (see [later in this section](#tags-decks-cards-and-notes) to search for tags). Some examples:
```

src/searching.md
brishtibheja:
```suggestion
mandatory, see [later in this section](#matching-special-characters).
```

brishtibheja:
@Danika-Dakika I have gone through all suggestions (I wished you added the suggestions in code block so I could just commit the changes directly; there's a button to do that when you comment), I have most of them resolved, otherwise I added comments. I have added a few extra commits too, probably the last ones. Hopefully, this will get merged when Dae comes. On a different note, are there sections of manual you deem important as a helper that haven't been updated in a while or is difficult-to-understand, etc.?

src/searching.md
Danika-Dakika: > Filtered decks don't give you a search result

They return a set of cards in the deck based on the filter, so I'd say they do give the user a search result. 

> do we really need to be technically correct?

Yes, that's pretty much the entire point of the manual.

src/searching.md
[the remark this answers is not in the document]
Danika-Dakika: I see how that's another fair reading of it.

src/searching.md
Danika-Dakika: > some amount of entropy is nice to have.

In the manual, or in any technical writing? -- I strongly disagree. This isn't prose. Parallel phrasing is how you quickly and efficiently communicate similarity and subtle differences between things. 

Adding variability makes things harder for a reader to understand at-a-glance, and ruins the effect of contrasting the examples in the first place. Different wording/terminology should be used to signal for a reader that the manual is talking about about something different -- because that's the impression it gives.

src/searching.md
Danika-Dakika: 👍 I'll try to remember that for next time.

src/searching.md
Danika-Dakika: I am not sure that clarifies things, especially since the reasoning runs counter to one of the primary concepts the manual needs to emphasize for users. 
> ... in Anki we add notes rather than cards, and Anki creates cards for us.

https://docs.ankiweb.net/editing.html#adding-cards-and-notes

src/searching.md
Danika-Dakika: Responded above at line 41.

src/searching.md
brishtibheja: We are already using the word in the manual and in the FSRS tutorial. I added the 'bar' but I think it shouldn't be an issue.

src/searching.md
brishtibheja:
```suggestion
finds notes without the exact sequence of characters "a dog" in them.
```

src/searching.md
brishtibheja: I have resolved the other conversation as the explanation itself was confusing. For this one, I want to ask @dae why the original phrasing was different for is:suspended and is:burued? The docs are created in your vision so I'd like to get your feedback first.

Danika-Dakika:
> On a different note, are there sections of manual you deem important as a helper that haven't been updated in a while or is difficult-to-understand, etc.?

It's an interesting question, but I don't want to think it through in this PR, because we'll never find it again. 😅 
I have a related idea that I'm going to post in the Forums, and it might lead to an answer for this!

src/searching.md
dae: I'm not sure this is exposed by AnkiMobile yet.

src/searching.md
```diff
@@ -183,148 +182,155 @@ Searches that ignore combining characters are slower than regular searches.
...
 
 `prop:due=-1`\
-cards due yesterday that haven’t been answered yet
+cards due yesterday that haven’t been answered yet.
+
+`prop:due>1`\
+all cards due in the future, including tomorrow.
+
+`prop:due<-1`\
+all cards due in the past (that haven't been answered yet), including yesterday.
```
dae, line 294:
```suggestion
all  overdue cards
```

dae (commented):
Thank you both for your hard work on this!

src/searching.md
dae: I can't remember any specific reason for the differing wording, and don't have strong feelings here.

dae:
`Custom data` allows storing extra data in cards. The old FSRS implementation used it, but it was not FSRS-specific, and may be used in niche other use-cases, but is probably obscure enough that we can just stop documenting it for now.

src/searching.md
brishtibheja:
```suggestion
of searching for specific cards/notes.  This method can also be used to adjust the scope of FSRS optimization. 
```

src/searching.md
brishtibheja:
```suggestion
Requires Anki 23.10+, AnkiMobile 23.10+ or AnkiDroid 2.17+.
```

src/searching.md
```diff
@@ -183,148 +182,155 @@ Searches that ignore combining characters are slower than regular searches.
...
-in a specific field, regular expressions in fields don't require an exact match. Eg:
+in a specific field, regular expressions in fields don't require an exact match:
 
 `front:re:[a-c]1`\
-matches uppercase or lowercase a1, B1 or c1 that occurs anywhere in the "Front" field
+matches uppercase or lowercase a1, B1 or c1 that occurs anywhere in the "Front" field.
 
 `front:re:^[a-c]1$`\
-like the above, but will not match if any other text falls before or after a1/b1/c1.
+same as previous, but will not match if any other text falls before or after a1/b1/c1.
```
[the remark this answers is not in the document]
brishtibheja:
```suggestion
same as the previous example, but will not match if any other text falls before or after a1/b1/c1.
```

brishtibheja:
@dae I believe this can be merged. I have also added documentation for escaping "and" and "or" so the other PR can be closed.

src/searching.md
```diff
@@ -376,12 +376,16 @@ If you're using a version earlier than Anki 2.1.36 the following searches may no
 As shown in the previous section, some characters like `*`, `_` and `"` have a
 special meaning in search. If you need to locate those characters in a search,
 you need to tell Anki not to treat them specially. This is called "escaping a character" and is primarily done by using double quotes and backslashes.
-
+
```
dae, line 379: Intentional?

brishtibheja: It's not.

src/searching.md
```diff
@@ -183,148 +182,155 @@ Searches that ignore combining characters are slower than regular searches.
...
 
 `prop:due=-1`\
-cards due yesterday that haven’t been answered yet
+cards due yesterday that haven’t been answered yet.
+
+`prop:due>1`\
+all cards due in the future, including tomorrow.
+
+`prop:due<-1`\
+all  overdue cards
```
Danika-Dakika, line 294: This just caught my eye as I was catching up on comments. 

```suggestion
`prop:due>=1`\
all cards due in the future, including tomorrow.

`prop:due<=-1`\
all  overdue cards
```

src/searching.md
dae:
```suggestion

```

dae:
Thank you both!

PR #221, opened 2024-07-30
Add special handling of "or" to searching

t-nil:
Just a little side fact that confused me


brishtibheja:
Hey, I'm updating this section in #219, and I can include this example in the sub-section about escaping characters if that's necessary. Although I believe we have enough examples there. Adding this here might not be a good decision.

dae:
I think the point here is how to search for a literal 'or' instead of two separate terms - might be worth adding in #219 if we don't already mention it

PR #227, opened 2024-08-05
added description for <meta>

brishtibheja:
The text used in `<meta name="description">` is often used by Google to generate search snippets and to provide more accurate search results.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -1,6 +1,6 @@
 [book]
 authors = ["Ankitects Pty Ltd and contributors"]
-description = "The official Manual for Anki, the flashcard program that makes learning easier."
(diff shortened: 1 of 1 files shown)


book.toml
```diff
@@ -1,5 +1,6 @@
 [book]
 authors = ["Ankitects Pty Ltd and contributors"]
+description = "The official Manual for Anki, the flashcard program that makes learning easier."
```
dae, line 3: `Manual`'s captitalization is odd.

How about this instead:

```suggestion
description = "Anki's user manual. Anki is a flashcard program that makes learning easier."
```

PR #232, opened 2024-08-09
Edit text / add images to Cards section of getting-started.md for clarity

dddevJACL:
In order to help new anki users, images were added to the Cards section in getting-started.md so that users can more directly connect what they are reading about to what they might see in anki. Some minor portions of the text in the Cards and Types of Cards section were also edited:

- Small sentence edits for clarity
- Mention that questions in anki can be plain text, but can also use media files, giving users more of a sense of the many ways anki can be used up front
- Identify the colors associated with each type of card in the Types of Cards section

If the images are appreciated, I can go through the getting-started.md file and add other images where relevant. I wanted to keep the scope of this pr small and focused on just one section however.


brishtibheja:
Most of this will conflict with my PR: #231

About this PR, first of all we are already overdocumenting. This section is supposed to give people a quick idea of how things work.

Second, telling people about the colours don't help as most of our readers are reading the basics and are not familiar with the UI. This should be reserved for the other sections.

Third, the images are a bad decision. No matter the amount of alt text, we will seriously impact our screen reader users and also might unintentionally harm other low vision readers. This is also hard to maintain and makes translation work more cumbersome as the translators will now have to get their own images. Images are also a distraction and in technical writing you will generally want to avoid it if you don't truly need to show people something.

dae:
Thank you for taking the time to contribute!

Re the colors, I feel less strongly about this than @brishtibheja, as the colors are used in multiple places. One alternative would be to show new/learning/review labels with colors instead of bold, instead of explicitly stating the color in an extra sentence, at least if we can do so in a way that looks fine in both light and dark themes.

Images can be useful when identifying particular parts of the UI (e.g. https://docs.ankiweb.net/browsing.html#search-tool), and things like diagrams to better explain concepts like cards vs notes would probably add a decent amount of value. But for the question/answer screenshots, I'm afraid I don't really think they're worth it.

brishtibheja:
The colours can be explained in studying.md, here I believe it will be a mere distraction.

PR #235, opened 2024-08-10
fixed a link

brishtibheja:
@dae This link was broken in the previous PR.


dae:
Thanks!

(side thought for the future: backticks might actually be a better choice than double quotes in some/all of these cases)

PR #236, opened 2024-08-10
minor fixes

brishtibheja:
I was about to add another commit but you merged the PR.

### DIFF
--- a/src/searching.md
+++ b/src/searching.md
@@ -64,25 +64,25 @@ Things to note from the above:
 - Search terms are separated by spaces.
 
 - When multiple search terms are provided, Anki looks for notes that
-  match all of the terms - an implicit "and" is inserted between each
+  match all of the terms - an implicit `and` is inserted between each
   term. On Anki 2.1.24+, AnkiMobile 2.0.60+, and AnkiDroid 2.17+ you can be explicit
-  if you like ("dog and cat" is the same as "dog cat"), but older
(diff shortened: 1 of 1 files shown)


src/searching.md
```diff
@@ -382,7 +382,7 @@ you need to tell Anki not to treat them specially. This is called "escaping a ch
   `part:"after the colon"`.
 
 - `And`/`Or`\
-  To search for the words `or` and `and`, wrap them with double quotes. For example, `dog "and" cat` searches for "dog", "cat" and the word "and".
+  To search for the words "or" and "and", wrap them with double quotes. For example, `dog "and" cat` searches for "dog", "cat" and the word "and".
```
dae, line 385:
```suggestion
  To search for the words `or` and `and`, wrap them with double quotes. For example, `dog "and" cat` searches for "dog", "cat" and the word "and".
```

brishtibheja: I understand your reasoning but double quotes are used as we're talking about the string of characters *to search* and not the search term itself even though we say "wrap them in double quotes".

dae: I'm afraid I find my suggested version easier to read than what you currently have.

Screenshot 2024-09-07 at 4 13 50 PM

brishtibheja: "to search for these words"? it's in a list, shouldn't confuse people.

dae: Works for me.

PR #245, opened 2024-08-16
Safe mode and video driver restart

Danika-Dakika:
- Added the confirmation message to instructions about safe mode (shift-start).
image
- Added instruction to restart after changing video drivers.

### DIFF
--- a/src/files.md
+++ b/src/files.md
@@ -64,7 +64,7 @@ syncing in order to use the full sync option without first downloading
 the changes. Similarly, if you are experiencing problems with Anki, you
 might need to (or might be instructed to) disable add-ons temporarily to
 see if one might be causing the problem. To do both of these things at the same time, you can 
-open Anki in safe mode by holding down the <kbd>Shift</kbd> key while starting, until the on-screen confirmation appears.
+open Anki in safe mode by holding down the <kbd>Shift</kbd> key while starting Anki. Keep holding <kbd>Shift</kbd> down until the on-screen message informs you that Anki has started in safe mode.
 
 It is possible to specify a custom folder location during startup. This
 is an advanced feature that is primarily intended to be used with
--- a/src/files.md
+++ b/src/files.md
@@ -64,7 +64,7 @@ syncing in order to use the full sync option without first downloading
 the changes. Similarly, if you are experiencing problems with Anki, you
 might want to (or might be instructed to) disable add-ons temporarily to
 see if one might be causing the problem. You can do both of these things
-by holding down the <kbd>Shift</kbd> key while starting Anki.
+by starting Anki in safe mode, holding down the <kbd>Shift</kbd> key until the on-screen confirmation appears.
 
 It is possible to specify a custom folder location during startup. This
 is an advanced feature that is primarily intended to be used with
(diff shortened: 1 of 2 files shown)


src/files.md
```diff
@@ -64,7 +64,7 @@ syncing in order to use the full sync option without first downloading
 the changes. Similarly, if you are experiencing problems with Anki, you
 might want to (or might be instructed to) disable add-ons temporarily to
 see if one might be causing the problem. You can do both of these things
-by holding down the <kbd>Shift</kbd> key while starting Anki.
+by starting Anki in safe mode, holding down the <kbd>Shift</kbd> key until the on-screen confirmation appears.
```
brishtibheja, line 67: This could've been more like:

```
To do both of these, you can start Anki in safe mode by holding down the <kbd>Shift</kbd> key while opening the app.
```

src/troubleshooting.md
```diff
@@ -12,7 +12,7 @@ If the problem does not occur again, you can skip the following steps.
 
 **2. Check add-ons**
 
-Please close Anki, then start it again while holding down the shift key. If the problem goes away, that indicates an add-on is causing the problem. Remove any add-ons you don't need, and disable half of the others. If the problem continues, try the other half. Repeat
+Please close Anki, then start it again in safe mode, holding down the <kbd>Shift</kbd> key until the on-screen confirmation appears. If the problem goes away, that indicates an add-on is causing the problem. Remove any add-ons you don't need, and disable half of the others. If the problem continues, try the other half. Repeat
```
brishtibheja, line 15: Here too. I don't think the phrases are well connected, it feels wrong to me.

src/files.md
Danika-Dakika: The whole point was to come up with an instruction that _incorporated_ the confirmation. I also wanted the phrasing to match between the 2 sections. It was hard to find a place for both "start Anki" and "while starting Anki" without sounding too redundant. If you really think this needs a "while" phrase, how about this? 

> To do both of these things at the same time, you can open Anki in safe mode by holding down the <kbd>Shift</kbd> key while starting, until the on-screen confirmation appears.

> then open it in safe mode by holding down the <kbd>Shift</kbd> key while starting, until the on-screen confirmation appears.

brishtibheja: Go ahead. The first paragraph sounds nice.

src/files.md
```diff
@@ -62,9 +62,9 @@ If you have made a destructive change on one computer and have an
 undamaged copy on another computer, you may wish to start Anki without
 syncing in order to use the full sync option without first downloading
 the changes. Similarly, if you are experiencing problems with Anki, you
-might want to (or might be instructed to) disable add-ons temporarily to
-see if one might be causing the problem. You can do both of these things
-by holding down the <kbd>Shift</kbd> key while starting Anki.
+might need to (or might be instructed to) disable add-ons temporarily to
+see if one might be causing the problem. To do both of these things at the same time, you can 
+open Anki in safe mode by holding down the <kbd>Shift</kbd> key while starting, until the on-screen confirmation appears.
```
dae, line 67:
```suggestion
open Anki in safe mode by holding down the <kbd>Shift</kbd> key while starting Anki. Keep holding shift down until a message informs you that Anki has started in safe mode.
```

dae: Perhaps this is a bit more legible? If you agree, we might want to update the one below too.

Danika-Dakika: Yes, that's definitely better. Thanks!

dae:
Thanks!

PR #246, opened 2024-08-16
Remove outdated "counts" paragraph

### LINKED ISSUE #243: Filtered deck counts
Is this accurate? I feel like I've never seen it that way, but it's been a while since I've used a filtered deck to advance cards. 

> **Counts:** 
> In a filtered deck, reviews that were already due are displayed in the review count as normal. **_Learning cards and non-due reviews are counted in the new card count_**, due to how the underlying implementation works.

https://docs.ankiweb.net/filtered-decks.html#counts

If that's outdated information -- the rest of the paragraph is about how Anki reschedules cards that were reviewed early, which is addressed later in the page, so it's possible the whole paragraph can be deleted.

Danika-Dakika:
closes #243


dae:
Thanks!

PR #247, opened 2024-08-16
Replace the source of zh-CN translation

L-M-Sherlock:
The previous zh-CN translation is outdated and nobody maintains it. I make a new translation project and it has been deployed here: https://open-spaced-repetition.github.io/anki-manual-zh-CN/


dae:
Thanks Jarrett! I got an email from someone who wanted to work on the old translation the other day, so I'll let them know about this.

PR #248, opened 2024-08-17
Deck options - clarify 2 separate timers

Danika-Dakika:
Folks seem to get confused about these timers a lot. I think clarifying there are 2 separate timers will help. I'd like to also change the name of the options for the on-screen timer, but that might be excessive. 

Potential in-app updates: 
1. Header of the deck options section: `Timer` → `Timers`
   * _I've already done this for the manual in this PR, so the questions are -- is this worth changing in the manual, and does it need to change in the app to match? If it doesn't make sense to do it, I'll revert it here_ 
2. Name of the option: `Show answer timer` → `Show on-screen timer`
3. Name of the option: `Stop timer on answer` → `Stop on-screen timer` (or `Stop on-screen timer on answer`)
   * _I didn't make these changes (2 & 3) yet for the manual, because changing feature-names is a bigger deal. Thoughts?_ 
4. [There's another string in anki/ftl/core/scheduling.ftl that is similar, but I'm not sure how it's related.]

Tooltip updates:

5. To match the text -- Once this language is locked, I will do a PR on anki/ftl/core/deck-config.ftl (I think that's the right spot) to add "-2" lines for: 
   * `On the review screen, show a timer that counts the time you're taking to study each card.`
   * `Whether to stop the on-screen timer when the answer is revealed. This doesn't affect statistics.`
6. If the header and option names are changed (1, 2, 3 above) [or are these just shared labels between the options and the tooltips?]


brishtibheja:
> I'd like to also change the name of the options for the on-screen timer, but that might be excessive.

It seems to me that readers of the manual are usually not as confused as other users are. I'm not against the changes you made, they'll certainly help some people. But might not the confusion be coming from what we use in the UI? This should be discussed in the forums. The rest of the points:

1. Maybe we can do better. Plurals aren't that consistent across languages, and sometimes translators even ignore them.
2. Good change.
3. Too long but still better.
4. Is this used somewhere? There are unused strings lying around and some are used in deck options but are coming from `scheduling.ftl`.
5. I don't think we add a "-2" to they key here.
6. The headers in markdown are only used as comments AFAIK. They don't necessarily need changing.

brishtibheja:
@Danika-Dakika In case you do some other changes in this file (while you're at it), I'm doing the whole section in #250, just remembered this one so will probably not touch "Timer".

Danika-Dakika:
> In case you do some other changes in this file (while you're at it)

I don't plan on expanding this PR beyond "Timer" -- and the list of collateral damage above,

dae:
- Proposed wording looks fine to me
- Someone will eventually point out any inconsistencies between manual and desktop, so we should only change one if we plan to change the other
- Creating new string with -2 will invalidate the existing translations, so they'll revert to English until a translator has a chance to translate them again. Has the English meaning changed enough that we need to force the translators to retranslate?

brishtibheja:
I am not so sure now. If a brght line distinction is made between them, one behaviour would be less intuitive: the on-screen timer stops at `Maximum answer seconds`. There seems to be some connection between them so they're not completely distinct, are they?

Danika-Dakika:
.1. "Timers" in the plural is sort of the point of this clarification, so I don't see any reason not to change it here and in the Deck Opts window. But I agree that it shouldn't need additional translation, so I'll just change the main string. 

.2. I think this is worth changing. It's only changing "answer timer" to "on-screen timer" so I'll just change the main string. 

.3. Even though "Stop on-screen timer on answer" is long, it's not longer than some of the other switches above/below it in Deck Opts. It's only changing "timer" to "on-screen timer" so I'll just change the main string. 

.4. Yes, I think that scheduling.ftl is where that string is coming from, because I can't find it anywhere else. 

.5. Agreed, I'll change the main tooltip strings to match the new language in the manual.

I'll do a PR in the other repo for scheduling.ftl, deckconfig.ftl, and help-page.ts. I think that's everywhere these appear. 


> If a brght line distinction is made between them, one behaviour would be less intuitive: the on-screen timer stops at `Maximum answer seconds`. There seems to be some connection between them so they're not completely distinct, are they?

@brishtibheja I see what you mean. I added this "(This timer will stop when it reaches the Maximum answer seconds set for the internal timer.)"

dae:
Thank you both!

PR #266, opened 2024-09-14
Added some more searches

brishtibheja:


user1823:
What's the difference between the following two? It would be nice to explain it clearly in the manual.

- Support resched:x for searching for cards that were manually rescheduled in x days (thanks to Henrik).
- Support prop:rated/resched to search for rated/rescheduled cards over specific time periods (thanks to Henrik).

From https://changes.ankiweb.net/changes/2.1.40-49.html

dae:
resched was originally added to distinguish it from regular study: https://github.com/ankitects/anki/pull/918. I don't remember the specifics, and could be wrong, but I think that resched:x was added to distinguish from rated:x.

Issue #260, opened 2024-08-30
Create a FAQ for "How answer buttons work"


brishtibheja:
Alongside this: https://github.com/ankitects/anki-manual/issues/255, we should create a seperate FAQ entry for "How answer buttons work" and replace the current two topics with "Grading your cards". To list everything:

1. Current organisation is bad. If I search "How answer buttons work" this will not show up as a result. The reorganisation will make it more searchable through Google and elsewhere.
2. A seperate section can allow a bit more exposition, and maybe FSRS's working when it becomes defualt.
3. A seperate FAQ entry will make space for https://github.com/ankitects/anki-manual/issues/255 which is more useful for most newbies reading the page. Manual should be more task oriented and keep unnecessary details out of it.
4. We can still link to the FAQ from manual wherever needed (in the learning steps topic, for example).

Expertium:
I agree. I propose this:
Easy - you answered correctly with no hesitation.
Good - you answered correctly with a little bit of hesitation.
Hard - you made a minor (not very important) mistake and/or you hesitated a lot.
Again - your answer was completely incorrect.

brishtibheja:
To be clear, my idea is having a "How answer buttons work" with technical explanations in a FAQ. In "Studying" section of manual, we should instead have "How to use answer buttons" like the one you describe. Another point that I didn't mention in OP is: when you have information unrelated to a task in Studying, a newbie may not know that "how to grade cards" doesn't depend on "how the buttons work". It's important to save people's time and not give them information that they later realise isn't much usable.

Expertium:
> To be clear, my idea is having a "How answer buttons work" with technical explanations in a FAQ

Done in https://github.com/ankitects/faqs/pull/34

dae:
I prefer not to talk about hesitation or recall difficulty when describing the answer buttons, as it's too subjective. How about something like:

- Use Again when you partly or completely forgot.
- Use Good when you answered correctly.

Those two buttons will typically be about 10% and 85% of your button presses.

The Hard and Easy buttons are useful for cards that are atypically hard or easy.

- Use Hard when you remembered correctly, and also want Anki to increase the delays for this card more slowly in the future.
- Use Easy when you remembered correctly, and also want Anki to increase the delays for this card more quickly in the future.

Expertium:
Those descriptions for Hard and Easy are too long, but more importantly, they force the user to think about intervals and not about how well they know what's written in the card.

brishtibheja:
[...]

PR #269, opened 2024-09-17
Added recommendations on how to use the answer buttons

Expertium:
Closes https://github.com/ankitects/anki-manual/issues/260

### DIFF
--- a/src/studying.md
+++ b/src/studying.md
@@ -147,7 +147,9 @@ a fail outside of Anki, then it counts as a fail in Anki as well.
 
 - **Easy** should be pressed if your answer is correct and it took no mental effort to recall it.
 
-Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct.
+Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct. The **Hard** and **Easy** buttons are useful for cards that are atypically hard or easy.
+
+If you don't feel confident in your ability to decide whether a card is harder/easier than other cards, you can only use Again and Good, which is a perfectly valid way of using Anki.
 
 ## Due Counts
 
--- a/src/studying.md
+++ b/src/studying.md
@@ -147,6 +147,8 @@ a fail outside of Anki, then it counts as a fail in Anki as well.
 
 - **Easy** should be pressed if your answer is correct and it took no mental effort to recall it.
 
+Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct.
+
 ## Due Counts
 
 When only the question is shown, Anki shows three numbers like 6 + 9 + 59
--- a/src/studying.md
+++ b/src/studying.md
@@ -141,7 +141,7 @@ without thinking about the algorithm.
 - **Again** should be pressed if your answer is incorrect. If your answer is partially correct, you should be strict with yourself: if it would count as
 a fail outside of Anki, then it counts as a fail in Anki as well.
 
-- **Hard** should be pressed if you weren't sure that your answer is correct or if you made a minor mistake. If your answer is 90% correct, press **Hard**.
+- **Hard** should be pressed if your answer is correct, but you had doubts about it or if you made a minor mistake. If your answer is 90% correct, press **Hard**.
 
 - **Good** should be pressed if your answer is correct, but it took some mental effort to recall it.
 
--- a/src/studying.md
+++ b/src/studying.md
@@ -133,6 +133,20 @@ answer. Pressing the <kbd>spacebar</kbd> or <kbd>Enter</kbd> will select **Good*
 See [Deck Options](deck-options.md) and the [FAQ](https://faqs.ankiweb.net/what-spaced-repetition-algorithm.html)
 to learn more about how the algorithm works.
 
+## How to choose the answer button
+
+The sections above explain *what* the answer buttons do, but you may also want to know *how* to choose an answer button intuitively,
+without thinking about the algorithm.
+
+- **Again** should be pressed if your answer is incorrect. If your answer is partially correct, you should be strict with yourself: if it would count as
+a fail outside of Anki, then it counts as a fail in Anki as well.
+
+- **Hard** should be pressed if you hesitated a lot before answering or if you made a minor mistake. If your answer is 90% correct, press **Hard**.
+
+- **Good** should be pressed if your answer is correct, but it took some mental effort to recall it.
+
+- **Easy** should be pressed if your answer is correct and it took no mental effort to recall it.
+
 ## Due Counts
 
 When only the question is shown, Anki shows three numbers like 6 + 9 + 59
--- a/src/studying.md
+++ b/src/studying.md
@@ -135,21 +135,19 @@ to learn more about how the algorithm works.
 
 ## How to choose the answer button
 
-The sections above explain *what* the answer buttons do, but you may also want to know *how* to choose an answer button intuitively,
-without thinking about the algorithm.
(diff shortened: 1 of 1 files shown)


brishtibheja:
I was wanting to do this eventually, but in a different way:

1. Remove the topics about how grades are chosen.
2. Retain these paragraphs somewhere:
> See [Deck Options](deck-options.md) and the [FAQ](https://faqs.ankiweb.net/what-spaced-repetition-algorithm.html)
to learn more about how the algorithm works.
> You can use the <kbd>1</kbd>, <kbd>2</kbd>, <kbd>3</kbd> and <kbd>4</kbd> keys on your keyboard to select a particular
button, where <kbd>1</kbd> is **Again**. Pressing <kbd>Space</kbd> or <kbd>Enter</kbd> will select
**Good**.
3. Name the current topic "Grading cards".

Expertium:
> Retain these paragraphs somewhere

The problen is - where?

src/studying.md
```diff
@@ -133,6 +133,20 @@ answer. Pressing the <kbd>spacebar</kbd> or <kbd>Enter</kbd> will select **Good*
...
 
+## How to choose the answer button
+
+The sections above explain *what* the answer buttons do, but you may also want to know *how* to choose an answer button intuitively,
+without thinking about the algorithm.
+
+- **Again** should be pressed if your answer is incorrect. If your answer is partially correct, you should be strict with yourself: if it would count as
+a fail outside of Anki, then it counts as a fail in Anki as well.
+
+- **Hard** should be pressed if your answer is correct, but you had doubts about it or if you made a minor mistake. If your answer is 90% correct, press **Hard**.
```
brishtibheja, line 144: That's an Again. Minor mistake = mistake.

src/studying.md
```diff
@@ -133,6 +133,20 @@ answer. Pressing the <kbd>spacebar</kbd> or <kbd>Enter</kbd> will select **Good*
...
+
+- **Good** should be pressed if your answer is correct, but it took some mental effort to recall it.
```
brishtibheja, line 146: It doesn't communicate that Good should be used most of the time. I also think we should stick with whatever dae wrote, with little modifications.

src/studying.md
Expertium: No, that's Hard, that's why I said "minor"

src/studying.md
Expertium: Dae's recommendations for Hard and Easy aren't great. We shouldn't make the user think about interval lengths or the algorithm.

Expertium: I added a clarification that Good is supposed to be the most commonly used button.

src/studying.md
Expertium: Again tanks the memory stability. If I made a non-critical mistake, it doesn't make sense to decrease memory stability by 1-2 orders of magnitude. So Hard, which doesn't decrease the interval length, makes more sense for something that is far closer to "fully recalled" than to "fully forgotten".

@dae this might seem like it contradicts my "no references to interval lengths or anything algorithmic in order to make the guidelines beginner-friendly" rule. But that rule is only for what is **written**. We, the people writing the guidelines, are free to use any reasoning behind the scenes, so long as what the manual says in the end is beginner-friendly.

brishtibheja: That only makes sense if you assume the information to be recalled is a unitary memory. I can quite reasonably think of the situation as complete forgetting of one of the component required to arrive at a correct answer.

Also, the example I shared before: If I see 春日 I think はるひ which is the correct reading but I always seem to forget that another possible reading is かすが、if I think "ah, this is a minor lapse of memory" I'm never going to unlearn my habit. So, I think Again is the right choice.

dae:
I'm reasonably happy with what these changes are trying to convey, but feel the wording could still be better. For example, how about changing the start of the section to remove the current intro, and using something like:

* The Good button marks your answer as correct. You'll typically use this button about 90% of the time.
* The Again button marks your answer as incorrect. You'll typically use this button about 10% of the time.

Hard and Easy marks your answer as correct, and tells Anki to be more or less aggressive when scheduling that card in the future. This can result in more efficient scheduling for cards that are uncommonly hard or easy, but some users feel that the extra time taken to decide on an answer button is not worth it. If you decide to use these buttons, you should expect to be using them about 5-10% of the time. If you find yourself marking the majority of your cards as Hard, you should probably be using the Good button instead.

Expertium:
1) Ok, no intro.
2) I disagree about percentages, since they imply a fixed retention rate. Let's not use percentages.
3) No "tells Anki to be more or less aggressive when scheduling" or "more efficient scheduling". Dae, please understand, this has to be as beginner-friendly as possible. Imagine that you are trying to explain this stuff to your grandma.

dae:
The reason why I think we might want percentages is they give users a ballpark figure for what's reasonable - if they find themselves using Hard 50% of the time, that's clearly a lot different to 5%. I'm aware there'll be some individual variation, but still feel that 'about 90%' and 'about 5%' might be more useful than providing no guidance whatsoever.

> Dae, please understand, this has to be as beginner-friendly as possible. Imagine that you are trying to explain this stuff to your grandma.

The tone I was going for was 'you don't need to use these buttons, but if you wish to, here's what they do'. I don't think you can avoid talking about what they do completely, but perhaps this could be reworded so such users don't feel the need to read further.

Expertium:
Alright, how about this?

> Again should be pressed if your answer is incorrect. If your answer is partially correct, you should be strict with yourself: if it would count as a fail outside of Anki, then it counts as a fail in Anki as well. You'll typically use this button about 5-15% of the time.
> Hard should be pressed if your answer is correct, but you had doubts about it or if you made a minor, non-critical mistake.
> Good should be pressed if your answer is correct, but it took some mental effort to recall it. When Anki is used properly, this should be the most commonly used button. You'll typically use this button about 80-95% of the time.
> Easy should be pressed if your answer is correct and it took no mental effort to recall it.

> Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct. If you don't feel confident in your ability to decide precisely how difficult a card is, you can only use Again and Good, which is a perfectly valid way of using Anki.

dae:
I'm ok with that wording. I personally think listing hard/easy further below and/or mentioning 2 buttons are valid further above might also be worth doing, but it's not a blocker.

dae:
> Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct.

This still leaves open questions about 'how incorrect is incorrect', and might risk encouraging users to use hard on partially-forgotten material.

Expertium:
Listing Hard and Easy separately would be weird.

> This still leaves open questions about 'how incorrect is incorrect'

I'd love to write an entire paragraph about 'how incorrect is incorrect', but the whole purpose of this section is to be clear and easy to understand. What wording would you use?

dae:
Perhaps it would be better not to talk about minor/non-critical mistakes? E.g.


---

> Hard should be pressed if your answer is correct, but you had doubts about it or if you made a minor, non-critical mistake.

Hard should be pressed if your answer is correct, but took a long time to recall.

> Simply put, press Again if your answer is completely incorrect, and press Hard, Good, or Easy if your answer is correct. 

Simply put, press Again if your answer was not completely correct, and press Hard, Good, or Easy otherwise. 

---
I think if we mention them, it might encourage the use of Hard when material has not been properly recalled.

dae:
Thank you!

brishtibheja:
Any issues with removing algorithmic information from here and link people to [here](https://faqs.ankiweb.net/what-spaced-repetition-algorithm.html#review-cards) and [here](https://faqs.ankiweb.net/what-spaced-repetition-algorithm.html#review-cards) instead ?

Explaining this doesn't make sense if people don't read deck options section.

PR #273, opened 2024-09-22
Adjust the percentages

Expertium:
https://www.reddit.com/r/Anki/comments/1fmtho5/comment/loddsn6/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

I knew adding percentages is not a good idea


brishtibheja:
I have 77 percent DR maybe in one of my presets. Such a large range is going to confuse people though :/

PR #285, opened 2024-10-04
Move FAQ info to Manual

brishtibheja:
@dae I've tried to include it in an existing file.


dae:
Thanks for all your work on these!

PR #286, opened 2024-10-07
Added a "Short Summary" section right under FSRS

Expertium:
Rationale: most users skim rather than read everything, and a lot of users just want a TLDR. @brishtibheja @user1823 @dae feel free to change the wording


Expertium:
Well, opinions are split. Let's see what Dae says.

dae:
I don't have strong feelings here. The text itself is fairly easy to follow, but sorata's points are also valid.

Re: desired retention, the text says it's the most important setting, but offers no guidance on what the user should change it to, which I think does more harm than good. If we must mention it in an intro guide, I think it should be done in a way that doesn't give users FOMO/stress about what it should be set to. For example (and I'm not suggesting we use this exact wording):

- Desired retention tells Anki how much of your material you want to remember. The default is 90%, which offers a good balance of recall and workload. You can adjust increase or decrease your desired retention to adjust how much work you need to do each day. [link to relevant menu section]

dae:
I'm still a bit on the fence, but since effort has been put into this and there was no clear consensus, I'll merge this in. If others feel strongly that this is a bad change, follow-up discussion or PRs are welcome.

PR #287, opened 2024-10-07
Create a page for creating good cards

brishtibheja:
It's not yet complete, but I'm opening the PR now to get some feedback. I have tried to read multiple sources for this one and have included only what I felt was essentially.

@Expertium Anything I missed to add? I haven't included some pretty basic guidelines from Woz (like use Mnemonics, use IO cards, use personal context, etc.) Also, I'm pretty bad at writing introductions to topics, so what I wrote here is basically an initially draft. Suggestions are welcome.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -54,7 +54,10 @@
 - [Troubleshooting](troubleshooting.md)
 - [Sync Server](sync-server.md)
 - [Miscellaneous](misc.md)
+  
+____
 
+[Effective Use of Anki](effective-use.md)
 [Contributing](contrib.md)
 [FAQs](faqs.md)
 [Resources](resources.md)
--- a/src/effective-use.md
+++ b/src/effective-use.md
@@ -1,0 +1,128 @@
+
+# Effective Use of Anki
+
+Anki can be used in various ways. However, to use it effectively, we recommend following the guidelines in this section. The guidelines here apply to most people using Anki regardless of the subject they are learning.
+
+## General Advice
+
+* **Memorise only when you understand**
+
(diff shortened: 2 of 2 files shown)


Expertium:
> I haven't included some pretty basic guidelines from Woz (like use Mnemonics, use IO cards, use personal context, etc.)

Why not?

dae:
I haven't read through the text carefully yet - before I do that and offer corrections, I just wanted to double check that you didn't reference the SuperMemo docs while writing this - if you did, it could be argued that this is a derived work, and we may not be able to use it.

Regarding the title: this page is mainly about card creation rather than things like study options, so perhaps it might be better called something like "Creating Good Cards" like in your subheading?

brishtibheja:
> you didn't reference the SuperMemo docs while writing this

Reference as in referencing in the writing itself? No. Or do you mean looking at SM articles? I've gone through SM's article, Control Alt Backspace's article, Andy's and other things I found online.

dae:
If you read SM's articles before (or especially during) writing this, I don't think we can include it as-is. I think either we need to give some bullet points about things we want to cover (i.e. section headings) to someone who hasn't read the page and get them to write the text instead ("clean-room design"), or we should mention at the top that the page is loosely based on a `[SuperMemo page](link to article)`.

brishtibheja:
It'll be hard to find people who haven't been influenced by Wozniak. I have included a "Further reading" section at the end. I think we can say something like "The following resources were used in writing this section" and either list some of the things I read or add a link to resources section.

Expertium:
I'd rather we just link to the 20 rules instead of reinventing the wheel: https://super-memory.com/articles/20rules.htm

PR #290, opened 2024-10-17
Wording Changes

brishtibheja:
Made it clear that the top-level decks' limits do not override subdeck limits.

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -137,7 +137,7 @@ make the backlog worse.
 
 By default, the daily limits of a higher-level deck do not apply if you're studying from its subdecks.
 
-If this option is enabled, the limits start from the top-level decks instead, which can be useful if you wish to study individual subdecks, while enforcing a total limit on cards for the deck tree. Note that the limits set on each subdeck will still apply.
+However, if this option is enabled, the limits start from the top-level decks. This option can be useful if you wish to study individual subdecks, while enforcing a total limit on cards for the deck tree. Note that the limits set on each subdeck will still apply.
 
 ## New Cards
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -137,7 +137,7 @@ make the backlog worse.
 
 By default, the daily limits of a higher-level deck do not apply if you're studying from its subdecks. A parent deck can have a new card limit of 10 cards/day and its subdecks can have a new card limit of 20 cards/day. The limits set on parent deck do not affect the number of new cards you can study from its subdeck.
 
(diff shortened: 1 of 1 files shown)


Gardengul:
"This means studying from the subdecks will also count towards the parent deck's limits."
What does it mean?

Gardengul:
"Enabling this option for any preset will affect all decks and presets in your collection."
Why was it deleted?

brishtibheja:
> "This means studying from the subdecks will also count towards the parent deck's limits."
What does it mean?

If you study 5 cards from the subdeck, the parent's limit decreases by 5 too.

> "Enabling this option for any preset will affect all decks and presets in your collection."
Why was it deleted?

We don't say this elsewhere in this page. A better way to denote global option might be just write a paragraph at the beginning and then use the globe icon or similar with names.

Gardengul:
>If you study 5 cards from the subdeck, the parent's limit decreases by 5 too.
But this is how it works regardless of whether "Limits start from top" is enabled or not.

brishtibheja:
Thanks for pointing out. Fixed now.

src/deck-options.md
```diff
@@ -135,7 +135,11 @@ make the backlog worse.
 
 ### Limits Start From The Top
 
-By default, the daily limits of a higher-level deck do not apply if you're studying from its subdeck. However, if this option is enabled, the limits will start from the top-level decks instead, which can be useful if you wish to study individual subdecks, while enforcing a total limit on cards for the deck tree. Enabling this option for any preset will affect all decks and presets in your collection.
+By default, the daily limits of a higher-level deck do not apply if you're studying from its subdecks. A parent deck can have a new card limit of 10 cards/day and its subdecks can have a new card limit of 20 cards/day. The limits set on parent deck do not affect the number of new cards you can study from its subdeck.
```
dae, line 138: How about the following instead, which makes it clear it depends on the deck you select, not the deck a card is from.

```suggestion
By default, the daily limits of a higher-level deck do not apply if you select one of its subdecks. A parent deck can have a new card limit of 10 cards/day and its subdecks can have a new card limit of 20 cards/day. The limits set on parent deck do not affect the number of new cards you can study from its subdeck.
```

src/deck-options.md
```diff
@@ -135,7 +135,11 @@ make the backlog worse.
...
+
+When this option is enabled, the limits set on higher-level decks also apply to their subdecks. In the previous example, you will be able to study only 10 new cards from the subdecks instead of 20 new cards.
```
dae, line 140:
```suggestion
When this option is enabled, the limits set on higher-level decks also apply to their subdecks when a subdeck is selected. In the previous example, you will be able to study only 10 new cards from the subdecks instead of 20 new cards.
```

dae:
Thank you both!

PR #291, opened 2024-10-17
Add More Resources

brishtibheja:
Added some free, open source resources that you can use with Anki.

### DIFF
--- a/src/resources.md
+++ b/src/resources.md
@@ -24,4 +24,8 @@
 
 ## Templates
 
-- [Prettify](https://github.com/pranavdeshai/anki-prettify#instructions): Collection of customizable Anki flashcard templates with modern and clean themes.
+- [Prettify by Pranav Deshai](https://github.com/pranavdeshai/anki-prettify#instructions): Collection of customizable Anki flashcard templates with modern and clean themes.
+
+- [Programming Templates from Rob](https://github.com/badlydrawnrob/anki#quick-start): Templates created for learning how to code with Anki.
+
+- [Memrise Templates from Eltaurus](https://github.com/Eltaurus-Lt/Anki-Card-Templates?tab=readme-ov-file#-quick-start): This card template recreates the Memrise interface in Anki.
--- a/src/resources.md
+++ b/src/resources.md
@@ -15,3 +15,13 @@
 ## Guides
 
 - [Anki Mastery Course](https://www.theanking.com/anki-mastery-course): A paid-for guide focusing on Anki use in medical schools.
+
+## Tools
+
+- [Yomitan](https://yomitan.wiki/): Yomitan is a dictionary pop-up that integrates with Anki. You can use this to create Anki cards for langauge learning.
+
+- [Jidoujisho](https://github.com/arianneorpilla/jidoujisho?tab=readme-ov-file): An Android application that allows you to create Anki cards for langauge learning.
+
+## Templates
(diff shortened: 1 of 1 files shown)


brishtibheja:
I haven't done this here, but [ErreAnki](https://community.ankihub.net/t/wiki-erreanki-medicina-e-residencia-by-ankiclass/132947) has been suggested to me. Do we want to add a resource like this that's potentially also trying to sell something? It's helpful for Portuguese/Brazilian medical students (I was told).

dae:
I am worried that the more links we add, the more other people will want to add their links there too. I think it might be best if that section is reserved for learning about Anki, rather than providing a link to all thinks with any connection to Anki.

brishtibheja:
I was partly inspired by Dynalist on this. They maintain an [official list of resources](https://dynalist.io/d/CQ4V16tFhIJIfy-rgDK7r1Bp) for their app Obsidian. I thought doing that for Anki would be nice too after I saw an old comment of yours saying you're fine with useful links that aren't trying to sell anything.

I agree with you that the Anki manual shouldn't become a soapbox for promotion but I see value in at least adding tools made for Anki if not templates etc. Such tools aren't always easily discoverable for users and having them here means Anki users can easily find them and they wouldn't get lost in the sands of time (as [NameLessGO said](https://forums.ankiweb.net/t/please-include-a-document-that-talks-about-how-to-make-better-cards/10225/4?u=sorata)). And we can be very selective in what is added.

The two tools I added in this PR can be a very significant part of some people's workflow. I'm sure I wouldn't be using Anki that much for my language learning if I didn't find an easy way to create vocab flashcards from text I read. I think they are enabling me to use Anki in a particular way and that's quite significant.

With the same logic, I think having Eltaurus's Memrise template here is useful. They are more than decorative and can actually help transitioning people from Memrise to Anki.

dae:
Where is the official list of resources linked from? If we're going to have such a list, perhaps it would make more sense to have it on an external page or forum wiki post, so it can be more easily updated, and is more apparent that it's not part of the official documentation (even if we link to it in the docs).

Expertium:
That's a good idea

brishtibheja:
@dae what's the next step here? I'd seriously like having a proper resource docs even if semi-official. 

What do you think of doing it in a google docs? If you can create a doc, people can directly suggest anything they want through the docs. 

I personally don't think forums are the best place to do it, what do you think?

dae:
I don't have any experience collaborating with Google Docs. Does that require someone to approve the suggestions? Who will be doing that? What criteria will they use to decide what goes in and what stays out?

The forums have a wiki feature that allows wiki posts to be edited by people other than the original poster. A wiki somewhere else could be used instead, but I suggested the forums because it's not extra software that needs to be hosted or paid for, and it's free of ads.

brishtibheja:
> I don't have any experience collaborating with Google Docs. Does that require someone to approve the suggestions?

Not necessarily. If you want, you can give editor access to everyone with a link. Or you can give people only access to comments and the suggestions will need to be approved manually by someone. I think the latter is better as no bad-faith edits will go through unless approved

If you can create a empty doc and share the link here, I can do the rest of the work. Or do you want this in the forums?

dae:
If you are happy to volunteer to do the moderation, please feel free to create this document yourself, and then make a new PR which points this section of the manual to it.

PR #297, opened 2024-11-04
Added descriptions of FSRS-related charts + descriptions of new stats

Expertium:
Related forum topic: https://forums.ankiweb.net/t/make-some-anki-stats-hyperlinks/51245/4

### DIFF
--- a/src/stats.md
+++ b/src/stats.md
@@ -114,7 +114,6 @@ overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
 Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. 
-Daily load = 1/I1 + 1/I2 + 1/I3 +...+ 1/In, where In is the current interval of the n-th card. 
 If a card's interval is less than a day, it counts as 1. If you don't have a backlog, daily load should be 
 approximately equal to your number of due cards.
 
--- a/src/stats.md
+++ b/src/stats.md
@@ -113,9 +113,10 @@ Note that the forecast graph does not count reviews that are currently
 overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
-Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. 
-If a card's interval is less than a day, it counts as 1. If you don't have a backlog, daily load should be 
-approximately equal to your number of due cards.
+Daily load is an estimate of the average number of cards to be reviewed daily 
+(assuming review at the scheduled time with no backlog. 
+If you don't have a backlog, daily load should be approximately equal to 
+your number of due cards.
 
 **Calendar**
 This graph shows past card review activity. Hovering the mouse over a specific item
--- a/src/stats.md
+++ b/src/stats.md
@@ -170,7 +170,7 @@ after each review.
 **Card Retrievability**\
 Only shown if FSRS is enabled. "Retrievability" is synonymous with "probability of recall".
 
-Estimated total knowledge is the number of cards the user is expected to know today, 
+Estimated total knowledge is the total number of cards you are likely to currently remember, 
 calculated by multiplying average retrievability by the number of cards that have been 
 reviewed at least once.
 
--- a/src/stats.md
+++ b/src/stats.md
@@ -171,7 +171,7 @@ after each review.
 Only shown if FSRS is enabled. "Retrievability" is synonymous with "probability of recall".
 
 Estimated total knowledge is the number of cards the user is expected to know today, 
-calculated as the product of average retrievability and the number of cards that have been 
+calculated by multiplying average retrievability by the number of cards that have been 
 reviewed at least once.
 
 **Hourly Breakdown**\
--- a/src/stats.md
+++ b/src/stats.md
@@ -114,7 +114,6 @@ overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
 Daily load is an estimate of the average number of cards to be reviewed daily 
-(assuming review at the scheduled time with no backlog). 
 If you don't have a backlog, daily load should be approximately equal to 
 your number of due cards.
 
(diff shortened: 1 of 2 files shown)


src/stats.md
```diff
@@ -113,7 +113,10 @@ Note that the forecast graph does not count reviews that are currently
 overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
-Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. Daily load = 1/I1 + 1/I2 + 1/I3 +...+ 1/In, where In is the current interval of the n-th card. If a card's interval is less than a day, it counts as 1. If you don't have a backlog, it should be approximately equal to your number of due cards.
+Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. 
+Daily load = 1/I1 + 1/I2 + 1/I3 +...+ 1/In, where In is the current interval of the n-th card.
```
brishtibheja, line 117: We don't want math here, trust me.

src/stats.md
```diff
@@ -113,7 +113,10 @@ Note that the forecast graph does not count reviews that are currently
 overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
-Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. Daily load = 1/I1 + 1/I2 + 1/I3 +...+ 1/In, where In is the current interval of the n-th card. If a card's interval is less than a day, it counts as 1. If you don't have a backlog, it should be approximately equal to your number of due cards.
+Daily load is an estimate of the average number of cards to be reviewed daily (assuming review at the scheduled time with no backlog. 
+Daily load = 1/I1 + 1/I2 + 1/I3 +...+ 1/In, where In is the current interval of the n-th card. 
+If a card's interval is less than a day, it counts as 1. If you don't have a backlog, daily load should be
```
brishtibheja, line 118: This is unnecessary, don't you think. It can also be incorrect if you increase/decrease your daily limits suddenly or if you aren't consistent in studying.

brishtibheja (changes requested):

src/stats.md
Expertium: I don't mind removing the math, but I do think that this part is important.

Gardengul:
The formula was the most important part. The rest is just an attempt to interpret it.

Expertium:
@dae this PR is ready

src/stats.md
```diff
@@ -113,6 +113,11 @@ Note that the forecast graph does not count reviews that are currently
 overdue, so if you have a large backlog, the overdue cards will not be
 displayed.
 
+Daily load is an estimate of the average number of cards to be reviewed daily 
+(assuming review at the scheduled time with no backlog).
```
dae, line 117: Do we need the part in brackets, given the paragraph that follows?

```suggestion
```

src/stats.md
```diff
@@ -151,10 +156,25 @@ graph than other graphs: rather than changing which cards or period of
...
+**Card Ease**\
 This graph shows the number of cards that have a specific ease factor.
 Average ease of the selected deck / collection is also displayed here.
 
+**Card Stability**\
+Only shown if FSRS is enabled. Stability is defined as the amount of time required
+for the probability of recall to decrease from 100% to 90%.
+
+**Card Difficulty**\
+Only shown if FSRS is enabled. Difficulty determines how quickly the intervals grow
```
dae, line 168: How about the following?

```suggestion
Only shown if FSRS is enabled. Difficulty determines how quickly a card's interval grows
```

src/stats.md
```diff
@@ -151,10 +156,25 @@ graph than other graphs: rather than changing which cards or period of
...
+after each review.
+
+**Card Retrievability**\
+Only shown if FSRS is enabled. "Retrievability" is synonymous with "probability of recall".
+
+Estimated total knowledge is the number of cards the user is expected to know today,
```
dae, line 174: Expected can read as a requirement ("you are expected to attend the class"), so how about the following instead?

```suggestion
Estimated total knowledge is the number of cards the user is likely to know today, 
```

Other suggestions welcome.

src/stats.md
```diff
@@ -151,10 +156,25 @@ graph than other graphs: rather than changing which cards or period of
...
+
+**Card Difficulty**\
+Only shown if FSRS is enabled. Difficulty determines how quickly the intervals grow
+after each review.
+
+**Card Retrievability**\
+Only shown if FSRS is enabled. "Retrievability" is synonymous with "probability of recall".
+
+Estimated total knowledge is the number of cards the user is expected to know today, 
+calculated as the product of average retrievability and the number of cards that have been
```
dae, line 175: In slightly more layman's terms:

```suggestion
calculated by multiplying average retrievability and the number of cards that have been 
```

dae (commented):
Sorry for the delay. Just some minor suggestions:

src/stats.md
Expertium: I prefer expected in this case. @brishtibheja thoughts?

dae: How about 'predicted'?

Expertium: "Predicted to know today" sounds odd to me, granted I'm not a native English speaker

dae: Sounds fine to me, but happy for someone else to weigh in.

brishtibheja: I think we should use "you" instead of "the user".

As for the verb, "likely" sounds best for the reason dae mentioned. But if we were to use "expected", how about: 
```suggestion
Estimated total knowledge is the number of cards you can expect to know today, 
```

I would still prefer to use:
```suggestion
Estimated total knowledge is the number of cards you are likely to know today, 
```

Expertium: I'll wait for Dae

dae: Mild preference for 'likely'. One other thought while we're bike-shedding: to avoid any confusion with cards that happen to be due today, one other idea would be something like:

Estimated total knowledge is the number of cards you are likely to currently remember.

Expertium: Ok, check the latest commit

dae:
👍

PR #299, opened 2024-11-08
Move to all "Note type"

### LINKED ISSUE #237: Note type versus notetype
It was said in #200 that we should move to notetype from the current inconsistent usage of note type/notetype.

I think "note type" is much more clearer than the other alternative, which also happens to be harder to read. In getting-started.md, for example, we explain "note type" after having explained a "note", instead if we use "notetype" the user will more likely feel he's encountering some CS jargon. If notetype is to be used though, I reckon we will need to switch to cardtype too. Both of these will be bad changes in my opinion. I think we should instead use "card type" and "note type" everywhere.

brishtibheja:
* Close #237

We'll need to change a link in Anki's repo with this. I'll link a PR.


src/templates/errors.md
```diff
@@ -138,9 +138,9 @@ a normal notetype, but this was never a use case that was intended to be support
 If you want to combine cloze deletions with regular cards, you may find an add-on
 like the Closet add-on provides some alternatives.
 
-### No Cloze Filter on Cloze Notetype
+### No Cloze Filter on Cloze Note Type
```
dae, line 141: We can avoid breaking the link:

```suggestion
### No Cloze Filter on Cloze Note Type
<div id="no-cloze-filter-on-cloze-notetype" />
```

dae:
Thank you!

PR #304, opened 2024-11-19
Update Broken/Outdated Link

### LINKED ISSUE #303: Frequently used jumplinks to studying.md are broken
Recent changes to studying.md [#289] have broken jumplinks, like --
https://docs.ankiweb.net/studying.html#learningrelearning-cards
https://docs.ankiweb.net/studying.html#review-cards

Before removing/changing section/sub-section names, consideration should be given both to internal manual links that will be broken [see: https://github.com/ankitects/anki-manual/issues/289#issuecomment-2484379093], and to *frequently*-used help links that are now useless. For these 2 especially, I think they need to be re-anchored to something -- either material on this page or the equivalent in a new location.

brishtibheja:
* Fixes #303

___

One other thing that we can do to help users is put a div with `learning-relearning`/`review-cards` id in here: https://github.com/ankitects/anki-manual/blob/main/src/studying.md?plain=1#L79

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -144,7 +144,7 @@ This option can be useful if you wish to study individual subdecks, while enforc
 ## New Cards
 
 The options here only affect new cards and
-[learning cards](studying.md#learningrelearning-cards). Once a card
+[learning cards](getting-started.md#card-states). Once a card
 has graduated (that is, gone through all the learning steps), the options in this section no longer apply to the card.
 
 ### Learning Steps
--- a/src/filtered-decks.md
+++ b/src/filtered-decks.md
@@ -170,12 +170,10 @@ and the desired retention in the deck preset.
 
 ## Steps & Returning
 
-Please see the section on [learning](studying.md#learningrelearning-cards) as a reminder of how
-steps work.
-
 By default, Anki will use the steps of a card’s home deck. If a new card
 would normally be reviewed twice when being learnt, the same thing will
 happen when you study it in a filtered deck.
+To learn about how (re)learning steps work, see [Deck Options](deck-options.md#learning-steps) section.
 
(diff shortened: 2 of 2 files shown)


Expertium:
https://docs.ankiweb.net/studying.html#how-to-choose-the-answer-button is also broken, the new one is https://docs.ankiweb.net/studying.html#answer-buttons. Ideally, we should keep using the previous one. This is definitely in the "frequently used" category, since I link to it on Reddit

brishtibheja:
I figured there aren't that many of these lying around so we probably can just correct the link from here on.

Expertium:
So are you saying that we should use https://docs.ankiweb.net/studying.html#how-to-choose-the-answer-button or that we should use https://docs.ankiweb.net/studying.html#answer-buttons? I can't tell.
I think we should use the previous link to avoid breaks.

src/deck-options.md
```diff
@@ -144,7 +144,7 @@ This option can be useful if you wish to study individual subdecks, while enforc
 ## New Cards
 
 The options here only affect new cards and
-[learning cards](studying.md#learningrelearning-cards). Once a card
+[learning cards](getting-started.md#card-states). Once a card
```
Danika-Dakika, line 147: Is this the right link to substitute? It doesn't have the same information as the moved/removed section.

Danika-Dakika:
This is a partial fix -- but if this is proposed as a fix to #303, it needs to also fix jumplinks to the headers you removed.

src/deck-options.md
brishtibheja: That info was moved to this very topic.

dae:
(merging this in as it's an improvement, but will avoid closing the issue)

PR #305, opened 2024-11-22
24.11

Expertium:
@Danika-Dakika @brishtibheja this is a draft, feel free to contribute. Ideally, it should be merged on the same day as when Anki 24.11 comes out.

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -493,9 +493,8 @@ and you set **Additional new cards to simulate** to 50, the simulator will simul
 the maximum number of reviews per day.
 - **Maximum interval** controls the maximum interval length (in days).
 
-To make the simulation as realistic as possible, the simulator takes into account real memory states (difficulty, stability, retrievability) of your cards as opposed
-to assuming that you start from zero, which is the assumption in CMRR. It also uses your FSRS parameters and the value of desired retention, therefore 
-changing them will affect the simulation.
+To make the simulation as realistic as possible, the simulator takes into account the real memory states (difficulty, stability, retrievability) of your cards. 
+It also uses your FSRS parameters and the value of desired retention, therefore changing them will affect the simulation.
 
 #### Learning and Relearning Steps
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -489,7 +489,7 @@ controls whether the simulator should simulate more cards than what already exis
 and you set **Additional new cards to simulate** to 50, the simulator will simulate a total of 150 cards. **New cards/day** and **Maximum reviews/day** control
 how many new cards will be learned each day and the maximum number of reviews per day. **Maximum interval** controls the maximum interval length (in days).
 To make the simulation as realistic as possible, the simulator takes into account real memory states (difficulty, stability, retrievability) of your cards as opposed
-to assuming that you start from zero, which is the assumption in CMRR.
+to assuming that you start from zero, which is the assumption in CMRR. It also uses your value of desired retention, therefore changing it will affect the simulation.
 
 #### Learning and Relearning Steps
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -511,7 +511,7 @@ contribute to long-term memory, so your time is
 better spent on other cards or a shorter study session.
 
 In the latest version of Anki you can let FSRS control short-term scheduling by leaving the (re)learning steps field empty.
-Note that just because FSRS-5 *can* give you intervals shorter than one day doesn't necessarily mean that it *will*. Your **Again** interval can be one day long or even longer.
+Note that just because FSRS-5 *can* give you intervals shorter than one day doesn't necessarily mean that it *will*. Your **Again** interval can be one day long, or even longer.
 
 #### Add-On Compatibility
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -368,6 +368,12 @@ Requires Anki 23.12 or later. Auto Advance allows you to automatically take some
 time in **Seconds to show question for** and/or **Seconds to show answer for**. Then, in the
 review screen, use the Auto Advance action from the **More** button to start advancing.
 
+## Easy Days
+
+If you want to spend less time on Anki on some days of the week, such as Sundays, this feature can help you do that.
+After the optimal interval is calculated, it will be adjusted by a small amount to change the due date.
+Note that setting all days to "Reduced" or "Minimum" will result in the same workload as setting all days to "Normal".
+
 ## FSRS
 
 The [Free Spaced Repetition Scheduler (FSRS)](https://github.com/open-spaced-repetition/fsrs4anki/wiki/ABC-of-FSRS) is an alternative to Anki's legacy
@@ -491,6 +497,9 @@ We also recommend you keep the number of learning steps to a minimum. Evidence
 shows that repeating a card multiple times in a single day does not significantly
 contribute to long-term memory, so your time is
 better spent on other cards or a shorter study session.
+
+In the latest version of Anki you can let FSRS control short-term scheduling by leaving the (re)learning steps field empty.
+Note that just because FSRS-5 *can* give you intervals shorter than one day doesn't necessarily mean that it *will*. Your "Again" interval can be 1d or even longer.
 
 #### Add-On Compatibility
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -373,6 +373,7 @@ review screen, use the Auto Advance action from the **More** button to start adv
 If you want to spend less time on Anki on some days of the week, such as Sundays, this feature can help you do that.
 After the optimal interval is calculated, it will be adjusted by a small amount to change the due date.
 Note that setting all days to "Reduced" or "Minimum" will result in the same workload as setting all days to "Normal".
+This feature works with both FSRS and the legacy SM-2 algorithm.
 
 ## FSRS
 
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -484,7 +484,7 @@ Compute minimum recommended retention (CMRR) attempts to find the desired retent
 
 In the latest version of Anki, you can use the simulator to get an estimate of your workload,
 either in reviews per day or in minutes of studying per day.
-'Days to simulate' controls the duration of the simulated study history. Additional new cards to simulate 
+'Days to simulate' controls the duration of the simulated study history. 'Additional new cards to simulate' 
 controls whether the simulator should simulate more cards being created. For example, if you currently have 100 cards under this preset,
 and you set 'Additional new cards to simulate' to 50, the simulator will simulate a total of 150 cards. 'New cards/day' and 'Maximum reviews/day' control
 how many new cards will be learned each day and the maximum number of reviews per day. 'Maximum interval' controls the maximum interval length in days.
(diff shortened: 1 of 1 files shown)


src/deck-options.md
```diff
@@ -474,6 +480,16 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
 
 Compute minimum recommended retention (CMRR) attempts to find the desired retention value that leads to the most material learnt, in the least amount of time. The calculated number can serve as a reference when deciding what to set your desired retention to. You may wish to choose a higher desired retention, if you’re willing to trade more study time for a greater retention rate. However, setting your desired retention lower than the minimum is not recommended, as you'll spend more time studying than necessary, due to increased forgetting.
 
+### The Simulator
+
+In the latest version of Anki, you can use the simulator to get an estimate of your workload,
+either in reviews per day or in minutes of studying per day.
+'Days to simulate' controls the duration of the simulated study history. 'Additional new cards to simulate'
```
brishtibheja, line 487: I would say use bold for these and not single quotes.

src/deck-options.md
```diff
@@ -474,6 +480,16 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
 
 Compute minimum recommended retention (CMRR) attempts to find the desired retention value that leads to the most material learnt, in the least amount of time. The calculated number can serve as a reference when deciding what to set your desired retention to. You may wish to choose a higher desired retention, if you’re willing to trade more study time for a greater retention rate. However, setting your desired retention lower than the minimum is not recommended, as you'll spend more time studying than necessary, due to increased forgetting.
 
+### The Simulator
+
+In the latest version of Anki, you can use the simulator to get an estimate of your workload,
+either in reviews per day or in minutes of studying per day.
+'Days to simulate' controls the duration of the simulated study history. 'Additional new cards to simulate' 
+controls whether the simulator should simulate more cards being created. For example, if you currently have 100 cards under this preset,
```
brishtibheja, line 488: How about something like:

```suggestion
controls whether the simulator should simulate more cards than what already exists under this preset. For example, if you currently have 100 cards under this preset,
```

src/deck-options.md
```diff
@@ -474,6 +480,16 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
...
+and you set 'Additional new cards to simulate' to 50, the simulator will simulate a total of 150 cards. 'New cards/day' and 'Maximum reviews/day' control
+how many new cards will be learned each day and the maximum number of reviews per day. 'Maximum interval' controls the maximum interval length (in days).
+The simulator relies on your FSRS parameters and cannot be used if FSRS is disabled.
```
brishtibheja, line 491: It's already called FSRS simulator so do we need to tell this?

I think telling people what other settings in the preset affect the simulation is helpful, though.

brishtibheja (changes requested):

src/deck-options.md
```diff
@@ -474,6 +480,18 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
 
 Compute minimum recommended retention (CMRR) attempts to find the desired retention value that leads to the most material learnt, in the least amount of time. The calculated number can serve as a reference when deciding what to set your desired retention to. You may wish to choose a higher desired retention, if you’re willing to trade more study time for a greater retention rate. However, setting your desired retention lower than the minimum is not recommended, as you'll spend more time studying than necessary, due to increased forgetting.
 
+### The Simulator
+
+In the latest version of Anki, you can use the simulator to get an estimate of your workload,
```
dae, line 485: Everything in the manual applies to the latest version of Anki (except when it's out of date), so this prefix doesn't really seem to add value.

```suggestion
You can use the simulator to get an estimate of your workload,
```

src/deck-options.md
```diff
@@ -474,6 +480,18 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
...
+either in reviews per day or in minutes of studying per day.
+**Days to simulate** controls the duration of the simulated study history. **Additional new cards to simulate**
```
dae, line 487: These all run together in a paragraph, and would be better separated into list elements like some other sections of this page.

src/deck-options.md
```diff
@@ -474,6 +480,18 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
 
 Compute minimum recommended retention (CMRR) attempts to find the desired retention value that leads to the most material learnt, in the least amount of time. The calculated number can serve as a reference when deciding what to set your desired retention to. You may wish to choose a higher desired retention, if you’re willing to trade more study time for a greater retention rate. However, setting your desired retention lower than the minimum is not recommended, as you'll spend more time studying than necessary, due to increased forgetting.
 
+### The Simulator
+
+In the latest version of Anki, you can use the simulator to get an estimate of your workload,
+either in reviews per day or in minutes of studying per day.
+**Days to simulate** controls the duration of the simulated study history. **Additional new cards to simulate** 
+controls whether the simulator should simulate more cards than what already exists under this preset. For example, if you currently have 100 cards under this preset,
```
dae, line 488:
```suggestion
controls whether the simulator should simulate more cards than this preset already has. For example, if you currently have 100 cards under this preset,
```

src/deck-options.md
```diff
@@ -474,6 +480,18 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
...
+and you set **Additional new cards to simulate** to 50, the simulator will simulate a total of 150 cards. **New cards/day** and **Maximum reviews/day** control
+how many new cards will be learned each day and the maximum number of reviews per day. **Maximum interval** controls the maximum interval length (in days).
+To make the simulation as realistic as possible, the simulator takes into account real memory states (difficulty, stability, retrievability) of your cards as opposed
```
dae, line 491:
```suggestion
To make the simulation as realistic as possible, the simulator takes into account the real memory state (difficulty, stability, retrievability) of your cards.
```

src/deck-options.md
```diff
@@ -474,6 +480,18 @@ If you're first switching from SM-2 and still wish to use this option, we recomm
...
+to assuming that you start from zero, which is the assumption in CMRR. It also uses your FSRS parameters and the value of desired retention, therefore
```
dae, line 492: Is this part worth saying?

```suggestion
It also uses your FSRS parameters and the value of desired retention, therefore 
```

src/deck-options.md
```diff
@@ -492,6 +510,9 @@ shows that repeating a card multiple times in a single day does not significantl
 contribute to long-term memory, so your time is
 better spent on other cards or a shorter study session.
 
+In the latest version of Anki you can let FSRS control short-term scheduling by leaving the (re)learning steps field empty.
```
dae, line 513:
```suggestion
You can let FSRS control short-term scheduling by leaving the (re)learning steps field empty.
```

dae: This is an experimental feature, so I think we should either clearly mark it as such, or avoid mentioning it yet.

src/deck-options.md
```diff
@@ -492,6 +510,9 @@ shows that repeating a card multiple times in a single day does not significantl
...
+Note that just because FSRS-5 *can* give you intervals shorter than one day doesn't necessarily mean that it *will*. Your **Again** interval can be one day long or even longer.
```
dae, line 514:
```suggestion
Note that just because FSRS-5 *can* give you intervals shorter than one day doesn't necessarily mean that it *will*. Your **Again** interval can be one day long, or even longer.
```

dae (commented):
Thank you for your efforts in keeping the manual up to date!

src/deck-options.md
Expertium: I think so

src/deck-options.md
Expertium: Each card has its own memory state, so the plural form (states) makes more sense, no?

src/deck-options.md
Expertium: I think we should mention it just to increase the number of people who use it, which could increase the rate at which bugs are discovered

src/deck-options.md
brishtibheja: I would favour dropping it. We should trim down fat from the docs. If you print it, it's like 200 pages which is too long.

src/deck-options.md
dae: I'm not a grammar expert, but to me it reads more naturally, as the word is attached to the singular 'memory', and sentences should read the same if parenthetical passages are removed.

dae: And having written that, I'm now second-guessing myself. If nobody agrees, feel free to leave it. :-)

src/deck-options.md
```diff
@@ -492,6 +514,9 @@ shows that repeating a card multiple times in a single day does not significantl
 contribute to long-term memory, so your time is
 better spent on other cards or a shorter study session.
 
+In the latest version of Anki you can let FSRS control short-term scheduling by leaving the (re)learning steps field empty.
```
dae, line 517:
```suggestion
In the latest version of Anki you can let FSRS control short-term scheduling by leaving the (re)learning steps field empty. This is an experimental feature.
```

src/deck-options.md
brishtibheja: It's about count noun versus mass noun. So, you wouldn't say states of affair because state is mass noun here. But, in our case I feel memory state is a countable thing and not an uncountable singular concept.

src/deck-options.md
```diff
@@ -368,6 +368,12 @@ Requires Anki 23.12 or later. Auto Advance allows you to automatically take some
 time in **Seconds to show question for** and/or **Seconds to show answer for**. Then, in the
 review screen, use the Auto Advance action from the **More** button to start advancing.
 
+## Easy Days
+
+If you want to spend less time on Anki on some days of the week, such as Sundays, this feature can help you do that.
+After the optimal interval is calculated, it will be adjusted by a small amount to change the due date.
+Note that setting all days to "Reduced" or "Minimum" will result in the same workload as setting all days to "Normal".
```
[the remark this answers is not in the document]
Expertium: 1 and 2. No, it's not FSRS-specific.
3. Yes, that's what it looks like.

src/deck-options.md
```diff
@@ -374,6 +374,7 @@ If you want to spend less time on Anki on some days of the week, such as Sundays
 After the optimal interval is calculated, it will be adjusted by a small amount to change the due date.
 Note that setting all days to "Reduced" or "Minimum" will result in the same workload as setting all days to "Normal".
 This feature works with both FSRS and the legacy SM-2 algorithm.
```
brishtibheja, line 376: Do we even need to document this?

Expertium: Yes, this is important. Some people might assume that it's FSRS-specific.

src/deck-options.md
```diff
@@ -374,6 +374,7 @@ If you want to spend less time on Anki on some days of the week, such as Sundays
 After the optimal interval is calculated, it will be adjusted by a small amount to change the due date.
```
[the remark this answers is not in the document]
Expertium: I removed it

src/deck-options.md
dae: I personally don't see the need, as the options screen automatically hides/shows options as FSRS is toggled. But I don't feel strongly about it.

PR #308, opened 2024-11-30
Move custom scheduling info from FAQ

brishtibheja:
Used info from the FAQ page to write about custom scheduling in the main manual.

Also, linked to FSRS's custom scheduling code as another example.


dae:
Thanks!

PR #311, opened 2024-12-01
Remove legacy import option

brishtibheja:
Legacy import code was removed in Anki 24.11 ver. https://github.com/ankitects/anki/pull/3536


dae:
On the one hand, I don't think we need to be in such a rush to remove these docs, as it'll take a while until the bulk of the user base has updated. On the other hand, if we don't do it around release time, it might get forgotten about.

Issue #313, opened 2024-12-03
Number of Cloze is limited in Nested Cloze


brishtibheja:
Nests more than 3 isn't allowed in latest (24.11) Anki ver. 

This needs to be added to `editing.md`.
https://github.com/ankitects/anki-manual/blob/bb51e749f4f33ac8395693ae964e5ee9d088b97a/src/editing.md?plain=1#L323
Relevant PR: https://github.com/ankitects/anki/pull/3264

PR #315, opened 2024-12-03
Add nested cloze limit to manual

GithubAnon0000:
Fix: #313

I previewed with `mdbook serve --open` as outlined in [CONTRIBUTING.md](https://github.com/ankitects/anki-manual/blob/main/CONTRIBUTING.md) and it works good on my end.

### DIFF
--- a/src/editing.md
+++ b/src/editing.md
@@ -402,8 +402,7 @@ The inner cloze is entirely nested within the outer. There is no support for par
 with the word "was" appearing in both deletions.
 
 In version 24.11 nested cloze deletions have been limited to a maximum of three levels.
-This change had been introduced in [PR #3264](https://github.com/ankitects/anki/pull/3264) to prevent
-slowdowns to the point of crashing anki.
+This change had been introduced to prevent slowdowns to the point of crashing anki.
 
 Prior to version 2.1.56, if you need to create clozes from overlapping text, add another Text
 field to your cloze, add it to the [template](templates/intro.md), and then when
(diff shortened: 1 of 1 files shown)


src/editing.md
```diff
@@ -401,6 +401,10 @@ The inner cloze is entirely nested within the outer. There is no support for par
 
 with the word "was" appearing in both deletions.
 
+In version 24.11 nested cloze deletions have been limited to a maximum of three levels.
+This change had been introduced in [PR #3264](https://github.com/ankitects/anki/pull/3264) to prevent
```
brishtibheja, line 405: Not sure about this. The docs don't need to mention the pull request as it's mainly meant for end-user consumption. We don't do this anywhere else.

brishtibheja: Yes, that's fine.

dae:
Thank you for contributing! I suggest we hold off on this for now, as the limit will either be bumped or made configurable in the next update.

dae:
The new limit is close to the old one. Do we need to document it anymore?

brishtibheja:
Most people search on a search engine and hopefully discussions on forums and here appear in the results. Other than that, troubleshooting already is asking people to update app.

Danika-Dakika:
I like the idea of documenting this *somewhere*, because this is the sort of thing that seems like we'll never forget it, but in 6 months it won't even sound familiar. 🤭

Perhaps a robust update on a well-named thread in the Forums would be enough?

dae:
Having given it some more thought, let's put it in the manual. Better that users are aware of the limit before they hit it.

dae:
Updated wording ok?

PR #319, opened 2024-12-05
Add Style Guidelines

### LINKED ISSUE #234: Make quotation usage consistent
I suggest we replace all instances of single quotations with either double quotes or *italics*. In some cases, single quotes are being used for emphasis where using *italics* might be better suited. In other cases, double quotes are better suited as this is what we use most of the time.

#231

### LINKED ISSUE #253: Use Bold Formatting for UI Elements
Manual doesn't have a consistent style when it comes to formatting, especially with UI elements where bold, code, quotes and also no formatting has been tried.

I believe we should use bold as most style guides recommend. In particular, [Google's way](https://developers.google.com/style/ui-elements).

### LINKED ISSUE #262: Double quotes vs backticks/kbd
Similar to the discussion on the searching page, you're changing backticks to double quotes here. Personally, I think backticks/kbd are more legible than double quotes, especially for things like example input. Thoughts?

_Originally posted by @dae in https://github.com/ankitects/anki-manual/pull/239#discussion_r1748021216_

brishtibheja:
This PR adds some style guidelines to `README.md`, mainly those that were discussed in the issue tracker.

The open issues that talked about style should now get closed and style issues can be handled in a simultaneous fashion when we update any section.

* Close #262
* Close #234
* Close #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ submit will be licensed under the [CC BY-SA 4 license](https://creativecommons.o
 The following conventions are not consistently applied in the manual. If you find such text, feel free to change it.
 
 - Use double quotes, `"`, instead of single quotes, `'`.
-- For emphasis, when defining anything or using words as words, use italics, `_`.
+- For emphasis such as when defining a word, use italics, `_`.
+- Also use italics when referring to a word itself rather than what it means, e.g. "To search for either _dog_ or _cat_, use `dog or cat`".
 - When referring to UI elements, use bold formatting, `**`.
 - When referring to keyboard buttons, wrap them with `<kbd></kbd>`.
 - Use code formatting for example inputs by wrapping them with backticks, `` ` ``.
--- a/README.md
+++ b/README.md
@@ -12,3 +12,13 @@ Contributions are welcome! Please see the [issue tracker](https://github.com/ank
 
 Unless stated otherwise, any pull requests you
 submit will be licensed under the [CC BY-SA 4 license](https://creativecommons.org/licenses/by-sa/).
+
+## Style Guidelines
+
+The following conventions are not consistently applied in the manual. If you find such text, feel free to change it.
+
+- Use double quotes, `""`, instead of single quotes, `''`.
+- When defining words or using words as words, use italics, `_` for emphasis.
+- When referring to UI elements, use bold formatting, `**`.
+- When referring to keyboard buttons, wrap them with `<kbd></kbd>`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,24 @@
...
+
+Unless stated otherwise, any pull requests you
+submit will be licensed under the [CC BY-SA 4 license](https://creativecommons.org/licenses/by-sa/).
+
+## Style Guidelines
+
+The following conventions are not consistently applied in the manual. If you find such text, feel free to change it.
+
+- Use double quotes, `"`, instead of single quotes, `'`.
+- For emphasis, when defining anything or using words as words, use italics, `_`.
```
dae, line 21: Using words as words?

dae (commented):
Thank you! Just one question:

README.md
brishtibheja: If you're talking about a word rather than what the word refers to, it's called "using words as words".

dae: It's my first time hearing that! Can you provide an example other than defining a word where it would make sense? Would you be comfortable with altered wording like the following instead?

"For emphasis, such as when defining words, use italics ..."

brishtibheja: > Can you provide an example other than defining a word where it would make sense?

"To search for either _dog_ and _cat_, use `dog or cat`"

> Would you be comfortable with altered wording like the following instead?
> 
> "For emphasis, such as when defining words, use italics ..."

Sure. Do we keep the other suggestion too?

dae: Sorry, I'm not following. Do you mean "words on words"? I was suggesting we drop that, as I at least found it confusing, so I'm assuming some other people may too. Or do you mean something else?

brishtibheja: See [this](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/NoneoftheAbove/faq0012.html) if you're confused what "words as words" mean.

As for where this guideline will apply for example, if you're talking about searching for the word "dog" in Anki, instead of putting double quotes around it, you'd use _italics_. That's the example I gave earlier.

I can write this as a seperate point with an example if this was confusing.

dae: I understand what you're saying, I just think that "words as words" is not particularly clear. I'd prefer you use another example instead of that particular phrase, or wait for someone else to chime in to see if they disagree with me.

brishtibheja: Check the latest commit.

(I couldn't find anything better at that time so decided to go with that.)

dae: Thank you!

Issue #241, opened 2024-08-12
Inputting Foreign Characters


brishtibheja:
This section gives people incomplete information and for specific instructions for any language simply asks them to use Google. This is not much helpful I believe, and should be removed.

If not this, I'll at least want to replace "Foreign" with something more inclusive like "Non-English". But ideally, I would think this getting removed is better.

Danika-Dakika:
Link?

brishtibheja:
[This one](<https://docs.ankiweb.net/editing.html?highlight=fore#inputting-foreign-characters-and-accents>)

dae:
The section exists so we can link people to it when they ask "How can I type ___ language in Anki?". It may not be perfect, but it's better than nothing, so I don't think we should remove it. No objections to wording tweaks, or fleshing out the instructions.

PR #320, opened 2024-12-07
Changed "Foreign" to "Non-Latin" and added linux related info

GithubAnon0000:
Fix: #241 

Not sure if adding detailed info regarding input methods is a good idea. It depends heavily on the input methods you use, which also depends on the operating systems.

Currently we list a few sources and even provide a few search prompts to aid the user. From there, they'll probably be able to get the keyboard working or ask in an appropriate forum for their OS.

### DIFF
--- a/src/editing.md
+++ b/src/editing.md
@@ -488,7 +488,7 @@ can use. Of note:
 While reviewing IO Cards a "Toggle Masks" button will appear just below the image.
 This button will temporary clear all shapes of the note when using "Hide All, Guess One" mode.
 
-## Inputting Non-English Characters and Accents
+## Inputting Non-Latin Characters and Accents
 
 All modern computers have built-in support for typing accents and
 non-english characters, and multiple ways to go about it. The method we
--- a/src/editing.md
+++ b/src/editing.md
@@ -523,12 +523,12 @@ not cover them all here. For more information, please try searching
 on the internet for "input Japanese on a mac", "type Chinese on Windows 10", and
 so on.
 
-For linux it's best to look at the wiki pages of your distro, e.g.
-[arch linux](https://wiki.archlinux.org/title/Input_method) and
-[debian linux](https://wiki.debian.org/Keyboard#Modern_strategy).
+For Linux it's best to look at the wiki pages of your distro, e.g.
+[Arch Linux](https://wiki.archlinux.org/title/Input_method) and
+[Debian Linux](https://wiki.debian.org/Keyboard#Modern_strategy).
 As an example, `apt install ibus-anthy` on Debian allows you to type hiragana characters.
 
-### Right-to-left languages
+### Right-to-left Languages
 If you are learning a right-to-left language, there are lots of other
 things to consider. Please see [this page](http://dotancohen.com/howto/rtl_right_to_left.html)
 for more information.
--- a/src/editing.md
+++ b/src/editing.md
@@ -502,30 +502,38 @@ often be typed on a generic "international keyboard" layout. These work
 by typing the accent, then the character you want accented - e.g. an
 apostrophe (´) then the letter a (a) gives á.
 
-To add the international keyboard on Windows machines, please see
-<https://thegeekpage.com/how-to-add-us-international-keyboard-in-windows-10/>
+### Adding international keyboard layouts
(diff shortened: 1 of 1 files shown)


src/editing.md
```diff
@@ -131,7 +131,7 @@ reviewing; for that, please see [templates](templates/intro.md).
 
 - **Editing Font** allows you to customize the font and size used when
   editing notes. This is useful if you want to make unimportant
-  information smaller, or increase the size of foreign characters which
+  information smaller, or increase the size of non-english characters which
```
brishtibheja, line 134:
```suggestion
  information smaller, or increase the size of non-latin characters which
```

src/editing.md
```diff
@@ -488,10 +488,10 @@ can use. Of note:
 While reviewing IO Cards a "Toggle Masks" button will appear just below the image.
 This button will temporary clear all shapes of the note when using "Hide All, Guess One" mode.
 
-## Inputting Foreign Characters and Accents
+## Inputting Non-English Characters and Accents
```
brishtibheja, line 491:
```suggestion
## Inputting Non-Latin Characters and Accents
```

src/editing.md
```diff
@@ -488,10 +488,10 @@ can use. Of note:
...
 
 All modern computers have built-in support for typing accents and
-foreign characters, and multiple ways to go about it. The method we
+non-english characters, and multiple ways to go about it. The method we
```
brishtibheja, line 494:
```suggestion
non-latin characters, and multiple ways to go about it. The method we
```

src/editing.md
```diff
@@ -500,17 +500,21 @@ have their own layouts specific to that language.
...
+apostrophe (´) then the letter a (a) gives á.
 
 To add the international keyboard on Windows machines, please see
 <https://thegeekpage.com/how-to-add-us-international-keyboard-in-windows-10/>
 
 To add it on Macs, please see
 <http://www.macworld.com/article/1147039/os-x/accentinput.html>
 
+To add it on Linux, please see
+<https://wiki.archlinux.org/title/Input_method> or <https://wiki.debian.org/Keyboard#Modern_strategy>.
```
brishtibheja, line 512: Is this not explaining an IME which is different from what the other links are doing here?

We can organise the info better in this topic.

brishtibheja: Looks much better than before.

brishtibheja:
You should've directly added all my suggestions: it's not *non-latin*, it's *non-Latin*.

src/editing.md
```diff
@@ -500,23 +500,40 @@ have their own layouts specific to that language.
...
+- KDE Plasma: <https://userbase.kde.org/Tutorials/ComposeKey>
+
+### Adding keyboard layouts for specific languages
+Keyboards for specific languages are added in a similar way, but we can
 not cover them all here. For more information, please try searching
-Google for "input Japanese on a mac", "type Chinese on Windows 10", and
+on the internet for "input Japanese on a mac", "type Chinese on Windows 10", and
 so on.
 
+For linux it's best to look at the wiki pages of your distro, e.g.
```
dae, line 526:
```suggestion
For Linux it's best to look at the wiki pages of your distro, e.g.
```

src/editing.md
```diff
@@ -500,23 +500,40 @@ have their own layouts specific to that language.
...
+[arch linux](https://wiki.archlinux.org/title/Input_method) and
```
dae, line 527:
```suggestion
[Arch Linux](https://wiki.archlinux.org/title/Input_method) and
```

src/editing.md
```diff
@@ -500,23 +500,40 @@ have their own layouts specific to that language.
...
+[debian linux](https://wiki.debian.org/Keyboard#Modern_strategy).
```
dae, line 528:
```suggestion
[Debian Linux](https://wiki.debian.org/Keyboard#Modern_strategy).
```

src/editing.md
```diff
@@ -500,23 +500,40 @@ have their own layouts specific to that language.
...
+As an example, `apt install ibus-anthy` on Debian allows you to type hiragana characters.
+
+### Right-to-left languages
```
dae, line 531:
```suggestion
### Right-to-left Languages
```

dae:
Thank you both!

PR #325, opened 2024-12-27
Add a link to the syncserver Dockerfile

omarkohl:


brishtibheja:
I think it's better to just move the info here, though I'm not aware of the reasons this was not documented in end-user docs but in developer docs.

omarkohl:
I am happy to do that if you (or someone else) can confirm that this is desired :slightly_smiling_face:

brishtibheja:
Let's wait till we hear from dae.

dae:
Happy with the link - it keeps the docker stuff in one place, which is helpful as it's user-contributed.

Issue #327, opened 2025-01-03
Document What to Do When TR is Low


brishtibheja:
The [stats page](https://docs.ankiweb.net/stats.html#today) should document the TR table and:

1. Explain people that a single day's data can be noisy.
2. Low TR can mean they haven't optimised and they should.
3. They don't have different presets for different material.
4. They might be using [answer buttons](https://docs.ankiweb.net/studying.html#answer-buttons) dishonestly.

PR #339, opened 2025-02-04
Added a description of the True Retention table

Expertium:
Closes https://github.com/ankitects/anki-manual/issues/327

However, I'm not sure if the stats page is the best place for teaching users.


dae:
Thanks! Not really bothered about the placement - a follow-up could move it to a different section and link to it if you wished.

PR #344, opened 2025-02-06
Update background.md to reduce cognitive load

GithubAnon0000:
(Hopefully) Solves: https://forums.ankiweb.net/t/creation-of-new-landing-page/42555/211?u=anon_0000 by @david-allison:
> The currently linked page is unstructured information overload and user unfriendly, especially compared to the new LP

I mainly reordered things into smaller sections to decrease cognitive load but also sometimes rephrased a few things.

I did use the markdown for quotes two times, because I wanted some sort of visual indication that we're dealing with examples here. Not happy that I used the quotes thingy, but I'm happy with the result of that quote thingy:
anki

### DIFF
--- a/src/background.md
+++ b/src/background.md
@@ -2,11 +2,14 @@
 
 <!-- toc -->
 
+## What is Anki?
 Anki is a program which makes remembering things easy. Because it is a
 lot more efficient than traditional study methods, you can either
 greatly decrease your time spent studying, or greatly increase the
 amount you learn.
 
+
+## Who can benefit from using Anki?
 Anyone who needs to remember things in their daily life can benefit from
 Anki. Since it is content-agnostic and supports images, audio, videos
 and scientific markup, the possibilities are endless. For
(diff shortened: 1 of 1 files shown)


david-allison:
Do you have a screenshot of "above the fold"?

This is what I see currently:


And a user is going to bounce as soon as they see what above the fold has to offer

* The table of contents takes up half the page
* The list of examples is a third of the page when scrolled
* https://readability.ncldata.dev/ is showing me a reading age between 15 and 19

For something like "learn more" from the home page, you'd want:

----

Anki is more than just digital flashcards. Just 10 Anki reviews can lead to you remembering something for X (20+) years. We do this using:

* **Active Recall** - You remember better  when you need to think of an answer, rather than picking the right answer from a list
* **Spaced Repetition** - You remember better if you spread your reviews out over time.

Anki can be used to learn anything...

brishtibheja:
@GithubAnon0000 It's a term that comes from newspapers to mean "what you see before opening the fold". In digital context, it means whatever you see before you start scrolling.

david-allison:
I'm responding to the ping, so my comments refer to using this page as the link from "learn more" on the home page.

**Above the fold** is the content which is seen before scrolling. It's the most important, as it entices a user to read the remainder of the page. 

The new website has ~22 words above the fold on my screen. It makes good use of whitespace. This is 5x this amount.

----

After this change, almost 100% of this page is a table of contents on mobile. That's not helpful to a user

----

> > https://readability.ncldata.dev/ is showing me a reading age between 15 and 19

> Doesn't sound bad to me at all. 

I believe the average reading age in the US is 7th grade (12-13). This page exists to convince people to learn more about why they should download Anki. This makes it difficult for English as a second language speakers to make an informed decision.

I'm not sure how accurate the tool is, but it's flagged as difficult content. 

----

> So you basically want a very simple page with as little content possible? I guess that would need to be in another document, then. The current background page seems to give an overview in more than a few brief sentences (mostly because of the history, really. But that is a part of ankis background).

I think the user needs to know the salient points within a few seconds of reading. If we hook the user, they'll read for a bit. 

I don't think a user expects a history of spaced repetition when they want to download an app

It doesn't seem appropriate to link this page in its current form. I do feel that effort should be put into ensuring that the secondary button on the landing page meets it's goal

david-allison:
I agree that we probably can't morph this into a good 'Learn More' page, and efforts should be redirected elsewhere.

I will question the value of the TOC: this is a page where you likely want to read in full from top to bottom.

david-allison:
I'm at capacity right now [I've spent a couple hours on AnkiDroid today, and my "To Review" queue is still at 12] and likely won't have time to se this through

PR #345, opened 2025-02-11
Removed "Optimize all presets" from the preset section

Expertium:
@brishtibheja @Danika-Dakika not gonna lie, I haven't been thoroughly checking the manual, so your help would be appreciated. There's probably more stuff that needs to be cleaned up. Also might as well lump [this](https://forums.ankiweb.net/t/several-fsrs-related-suggestions-again/53309/36?u=expertium) together as well.

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -38,7 +38,6 @@ arrow icon in the top right of the Deck Options window. You can do the following
   a [one-way sync](./syncing.md#conflicts).
 - **Save to All Subdecks**: Like **Save**, but also assigns the selected preset to all
   subdecks of the currently selected deck.
-- **Optimize All Presets**: When FSRS is enabled, this allows you to optimize the parameters of all presets at once.
 
 Deck options are not retroactive. For example, if you change an option that
 controls the delay after failing a card, cards that you failed before


Danika-Dakika:
Hmmm... I think this setting was there for enough versions that we don't need to rush to strip it out the day after a new version is released. [While I also acknowledge that this is the sort of attitude that ends up with things getting forgotten ...] The manual needs to serve a wider audience than just 1st-day-updaters 😉 -- so it might be more prudent to add a version mention here. 

But if this is going to change -- what about updating it in the FSRS section down the page? 

> You can also optimize the parameters for all of your presets at once, by clicking on the down arrow in the top right corner of the deck options window, then choosing Optimize All Presets.

Expertium:
Alright, let's put it on hold for a few weeks or a month.

dae:
Please ping us in another two weeks. I agree with Danika though - for future changes, I think we might be better off by adding rather than removing, e.g. adding "From Anki 25.02, this option has moved into the FSRS section", which we can then clean up at a future date.

Expertium:
Unrelated, but I find "deck preset" to be confusing, we should probably reword that.

Expertium:
@dae

PR #353, opened 2025-03-07
Updated a line for better flow of words

PradyumnaFuke:
I tweaked it for better flow.


PradyumnaFuke:
I missed the question mark, maybe because there wasn't one already so it never came to my mind.

brishtibheja:
I agree. You can close the PR from this page.

PR #354, opened 2025-03-10
Adding/Editing: describe how to access fields of IO cards

Zocker1999NET:
as those are not intuitively to find for all, as the UI has no indication for that except for that really small "Toggle Mask Editor" button.

### DIFF
--- a/src/editing.md
+++ b/src/editing.md
@@ -496,6 +496,11 @@ can use. Of note:
 - Ungroup selection: Select a group and then click this button to make each shape independent again.
 - Alignment: This tool can be used to align your shapes / text areas as desired.
 
+<!-- fields & tags are not intuitive to find in editor -->
+By default, IO notes also have fields for an header, a text for the back
+and a comment field. To access those in the editor, click on "Toggle Mask Editor".
+There you may also edit the tags of IO notes.
+
 While reviewing IO Cards a "Toggle Masks" button will appear just below the image.
 This button will temporary clear all shapes of the note when using "Hide All, Guess One" mode.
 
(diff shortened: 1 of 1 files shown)


src/editing.md
```diff
@@ -496,6 +496,11 @@ can use. Of note:
 - Ungroup selection: Select a group and then click this button to make each shape independent again.
 - Alignment: This tool can be used to align your shapes / text areas as desired.
 
+<!-- fields & tags are not intuitive to find in editor -->
+By default, IO notes also have fields for an header, a text for the back
+and a comment field. To access those in the editor, click on "Toggle Mask Editor".
+There you may also edit the tags of IO notes.
```
Danika-Dakika, line 502: Great idea! Thanks for noticing this. 

Some suggestions for the text -- 
```suggestion
The default IO note type also has standard fields: **Header** (displayed above the image on the front and back of each card), **Back Extra** (displayed below the image on the back of each card), and **Comments** (not displayed on the cards). To access those from the IO editor, click the **Toggle Mask Editor** button. There you can also view and edit the **Tags** of the note.
```

What do you think about moving this up to the "Adding" section? Between the sample image and the click-Add paragraph, maybe? 
image


I'm not certain why all of the editing features were separated into a different section in the first place since the same buttons are available when you are _adding_ an IO note. We might want to consider merging the sections in the future. For now, it makes sense to explain the fields and tags at the Add stage, because that's when you're most likely to be using them.

dae:
Do you plan to return to this?

Zocker1999NET:
sorry for taking so long to notice this.

@Danika-Dakika Thank you for your suggestions. I agree with you in those points & implemented them.

Danika-Dakika:
Thanks!

PR #355, opened 2025-03-14
Mention has-cd query

user1823:
Introduced in https://github.com/ankitects/anki/pull/2626

### DIFF
--- a/src/searching.md
+++ b/src/searching.md
@@ -313,6 +313,9 @@ cards that been lapsed more than 3 times.
 `prop:ease!=2.5`\
 cards easier or harder than default ease.
 
+`has-cd:v`\
+cards having the property `v` in custom data (requires Anki 23.10+).
+
 `prop:cdn:d>5`\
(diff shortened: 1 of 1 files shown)


Danika-Dakika (commented):
What do you think about moving this below the `cdn:d>5` example -- since it was introduced in a later version? That would match it up with the `cds:v=reschedule` example introduced at the same time. 

If we wanted to get really crazy -- could both of the 23.10 changes get moved below this line (if it's accurate that they would only be useful with FSRS enabled?), and then we don't need version information on every line? 

> The following searches require Anki 23.10+ and FSRS enabled:

user1823:
I chose the order based on their function, rather than the Anki version. It makes sense to mention that query first which is used to check whether a particular property is present in the custom data and then the queries that are used to select cards based on the values of those properties.

dae:
If you both are happy with the change I've made, please feel free to merge.

user1823:
LGTM now.

PR #358, opened 2025-03-22
Minor changes to awkward wording

Expertium:


Danika-Dakika:
It's amazing the things I look at almost every day and never see the actual words! Thanks!

brishtibheja:
@Expertium @Danika-Dakika The "which vs that" is a dialect issue. In UK+most commonwealth countries, the original wording would be preferable, although both are fine (it has to do with restrictive/non-restrictive clauses). Just a FYI, as I was the one who wrote that. I'm ofc fine with this change.

PR #359, opened 2025-03-24
Update wording of stats.md ("should" → "is expected to")

GithubAnon0000:
Also see: https://github.com/ankitects/anki/pull/3874#issuecomment-2747594117

originally reported by @Expertium.


brishtibheja:
FYI, google's recommendations on these verbs are quite good: https://developers.google.com/style/prescriptive-documentation#word-choice

Issue #340, opened 2025-02-04
Suggestion: Add the official apps to the Platform Notes and Getting Started section of the manual


Irrevenant2:
I suggest adding information about the official Android and iPhone apps to the Platform Notes and Getting Started section of the manual, please.

That might hopefully prevent some people from getting scammed by the unofficial for-pay apps, as well as including the apps availability and explaining how the apps interact with the other versions of the software.

Danika-Dakika:
The reason why that isn't there already is that this is the manual for the *desktop* version(s) of Anki. Even though a lot of it applies to all platforms, [AnkiDroid](https://docs.ankidroid.org/) (which is produced by a separate group, so just outside of "official") and [AnkiMobile](https://docs.ankimobile.net/intro.html) have their own manuals.

Irrevenant2:
Thanks.

1. The manual (assuming we're all talking about the same one at https://docs.ankiweb.net/) doesn't say that it's for the desktop version. It's just called 'Anki manual'. Should the name of the manual be clearer? I suspect for most people 'Anki' means the ecosystem, not specifically the desktop version.

2. If this manual is for the desktop version then I agree it doesn't make sense to list the app versions with the same weighting. IMO it would still be useful to include a line in the nature of a note at the bottom of https://docs.ankiweb.net/platform/intro.html saying (for example) 'There are also apps available at [AnkiDroid](https://play.google.com/store/apps/details?id=com.ichi2.anki&hl=en_AU) and [AnkiMobile](https://apps.apple.com/us/app/ankimobile-flashcards/id373493387)'.
 
3. It's good that this info is in the FAQ but FAQs are for answering user questions and this is an issue users mostly don't know that they need to ask about. They're unlikely to think "Is there an app version for this and are there a bunch of fake apps that I should avoid?". IMO it's worth a mention in the main manual body.

brishtibheja:
Not unreasonable to mention them in "Getting Started" IMO.

Irrevenant2:
> Yes, it is for the desktop version.

Thanks. The manual doesn't clearly state (other than on the platform notes page) that "Anki manual" is the manual for the Anki desktop program not the Anki ecosystem. A mention of that in the manual name or at least intro page might be a good idea.

>If people don't even read the short download instructions on the website, then they'll be much less likely to read the lenghty manual – that's my guess at least.

[...]

PR #361, opened 2025-03-29
Add official apps to manual

GithubAnon0000:
Fixes: #340 

This PR…
- changes the title to make it obvious that this is the manual for the desktop version of Anki
- includes the mention of other official apps copied from the [FAQ](https://faqs.ankiweb.net/anki-knockoffs.html).

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -4,7 +4,7 @@ description = "Anki's user manual. Anki is a flashcard program that makes learni
 language = "en"
 multilingual = false
 src = "src"
-title = "Anki Manual"
+title = "Anki Manual for Desktop"
 
 [build]
 create-missing = true
--- a/book.toml
+++ b/book.toml
@@ -4,7 +4,7 @@ description = "Anki's user manual. Anki is a flashcard program that makes learni
 language = "en"
 multilingual = false
 src = "src"
-title = "Anki Manual for Desktop"
+title = "Anki Manual"
 
 [build]
 create-missing = true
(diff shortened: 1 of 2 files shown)


brishtibheja:
I think we should avoid the word "desktop" as people with laptops etc. can also use this. Dae once said that this is a point of confusion for some people.

IMO "introduction" should mention which particular client app this manual is about, wdyt?

dae:
Mac users may not think 'PC' targets them. "Anki for computers" could work, but titles should be succinct, and I have a mild preference for the old title. What does everyone think?

dae:
Sorry, I'm afraid seeing the mobile screenshot changed my mild preference for the old title to a not-so-mild preference :-( I feel like it looks awkward, and can't think of another cross-platform app that uses such wording in their title. Is there anyone else who wants to weigh in on whether we should/shouldn't do this?

Rest of the PR looks good to go 👍

dae:
Yes, the title change is the only thing blocking this

dae:
Yep, all good - the changes get squashed together on merge. Thank you!

Irrevenant2:
"Desktop" is the common name for the style of GUI used by Windows etc. that you navigate using a mouse. It's called that whether you use it on a PC or a laptop. AFAICT it's the right word to use here.

Irrevenant2:
> "Desktop" is the common name for the style of GUI used by Windows etc. that you navigate using a mouse. It's called that whether you use it on a PC or a laptop. AFAICT it's the right word to use here.

Never mind, I was reading the comments in the wrong order. 'Anki for computers' seems fine.

PR #365, opened 2025-04-09
Update spelling in leeches.md - learnt to learned

DolofonosAtakura:
"Learnt" is okay when speaking (usually in UK English), but the spelling should be "learned"

### DIFF
--- a/src/leeches.md
+++ b/src/leeches.md
@@ -42,7 +42,7 @@ useful for future reference, you can leave it suspended.
 ## Waiting
 
 Some leeches are caused by "interference". For example, an English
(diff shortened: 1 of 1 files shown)


brishtibheja:
It's quite common to use "learnt" where I live (former colony, UK-like standards) but it has occurred at least once before when someone tried to "fix" this spelling. Maybe we should move to all "learned" as that's common everywhere? To me at least, it seems like some US people find "learnt" awkward to use.

dae:
I'm an Australian myself, but in recent years I've been trying to use American English in the manual and app. I don't recall a British English speaker ever trying to correct an Americanism in the manual or app, but the opposite has happened many times. 🙄

PR #367, opened 2025-04-10
Add faq (My interval went down (e.g. from 4 to 3 days) after optimizing my FSRS parameters. Why is that?)

GithubAnon0000:
`Fixes`: ankitects/anki#5455 
Depends on ankitects/anki-manual#403.

Please see https://github.com/ankitects/anki-manual/pull/367#issuecomment-2937156866 down below for images demonstrating the change.

<details><summary>Old description</summary>
<p>

# Changes
1. FAQ section has been moved up in the menu, because it didn't allow me to create subsection otherwise.
2. A subsection (FSRS) has been created for the FAQ.
3. The FAQ front page description has been updated.
4. An FAQ (My interval went down (e.g. from 4 to 3 days) after optimizing my FSRS parameters. Why is that?) has been added.

# Others
Regarding ankitects/anki#5466: I added it to the manual, but the way it currently is is quite awkward. Ideally, there'd be only one place to look for the FAQs.

I think it's best to move all FAQs from https://faqs.ankiweb.net/ into the FAQ section of the manual, if they haven't been covered already in the main parts of the manual. Afterwards, the link to the faq page should be removed from the faq.md file.

</p>
</details>


Expertium:
Oops, typo
Reterievability -> Retrievability

brishtibheja:
I don't like this extra subsection. Is there a issue with having this in the same page? There can be a horizontal line seperating the FAQ you're trying to add and the link to the soon-to-be-legacy FAQ site.

brishtibheja:
I have an idea for how a single-page FAQ section would look like and it doesn't have to have a ton of `<hr>`.

> Since it's very likely that the FAQ section will expand in the future

The thing is, if we remove this page later the link would stop working. If, however, we start with the existing page and move towards subsections if/when needed, the current links would still work in the future.

dae:
I'm afraid I'm not fond of this:

- FAQs are not where the manual should start from
- I think big FAQ lists are a bit of an anti-pattern. The info is best placed in an appropriate section of the manual, with some topic-specific mini-FAQs being placed on the page if they're really so common that it makes sense to call them out specifically.

dae:
If something makes sense to mention in the relevant section of the manual, I'm suggesting we mention it directly there, not create a separate FAQ page and link to it.

PR #369, opened 2025-04-15
Remove RMSE Explanation

### LINKED ISSUE #368: Update RMSE (bins) description
RMSE (bins) mathematical description: [The-Metric](https://github.com/open-spaced-repetition/fsrs4anki/wiki/The-Metric)
To make the example clear, suppose that there is only one bin.

The anki manual states:
> RMSE (bins) can be interpreted as the average difference between the predicted probability of recalling a card (R) and the actual probability measured from your review history. For example, RMSE=5% means that, on average, FSRS is off by 5% when predicting R. 

This is wrong. For example, if we flip a fair coin 4 times then the optimal predictions should be
`R1 = [0.5, 0.5, 0.5, 0.5]`. 
If we perform the coin flips and observe:
`[0, 1, 1, 0]`,
then RMSE (bins) is 0 for the optimal prediction, but the following erroneous prediction would also result in RMSE (bins) of 0: 
`R2 = [1, 0, 0, 1]`. 
The error of each prediction is `[1, 1, 1, 1]` since whenever R=0 is predicted we observe Heads, and whenever R=1 is predicted we observe Tails, so judging by the description of RMSE (bins) it would be more accurate to say that RMSE (bins) should result in 1.0. But here the RMSE (bins) formula wrongfully returns 0 since it only cares about the average of the predictions compared to the average of the observations:
`mean([0.5, 0.5, 0.5, 0.5]) = mean([1, 0, 0, 1]) = 0.5` which has the same value as the average of the observations `mean([0, 1, 1, 0]) = 0.5`.

There is an alternative interpretation of "the actual probability measured from your review history", so if we observed [0, 1, 1, 0] then we can suppose that the probability of heads for the coin is p=0.5. In this case, we get an error of [0.5, 0.5, 0.5, 0.5] from R2 which is still a non-zero average error.

A similar example is that for a sequence of n fair coin flips, the prediction
`R3 = [0, 1, 0, 1, ..., 0, 1]`
would usually perform better than
`R4 = [0.55, 0.55, 0.55, 0.55, ...]`
since `mean(R3) = 0.5` and `mean(R4) = 0.55`. As the number of coin flips goes to infinity, RMSE (bins) of R3 approaches 0.0 while RMSE (bins) of R4 approaches 0.05.

I suggest to add RMSE (bins) to this line and delete the existing description unless we can find an accurate interpretation for RMSE (bins):

> Log loss doesn't have an intuitive interpretation.

brishtibheja:
Fix #368

### DIFF
--- a/src/deck-options.md
+++ b/src/deck-options.md
@@ -457,13 +457,7 @@ down arrow in the top right corner of the deck options window, then choosing **O
 **Evaluate FSRS Parameters**
 
 You can use the **Evaluate** button to see metrics that show how well the parameters fit your review history. Smaller numbers
-indicate a better fit to your review history.
-
-Log loss doesn't have an intuitive interpretation. RMSE (bins) can be
-interpreted as the average difference between the predicted probability
-of recalling a card (R) and the actual
-probability measured from your review history. For example, RMSE=5% means that, on average, FSRS
-is off by 5% when predicting R. You don't need to understand these metrics to use FSRS.
+indicate a better fit to your review history. The metrics are hard to interpret intuitively so this is all you need to know.
 
 Note that log loss and RMSE (bins) are not perfectly correlated,
 so two decks may have similar RMSE values but very different log-loss values, or the other way around.


Danika-Dakika:
It's quite premature to PR this on an issue that is far from settled. Let's resolve what we're going to do about it first, and then you can write up the changes.

Danika-Dakika:
The new text was dismissive and propose in the midst of a discussion about what the feature even is and what it will be. Let's settle it first, and then we can document it.

PR #372, opened 2025-04-25
Remove CMRR

Expertium:
It's almost certain that CMRR will be removed in Anki 25.05 because it outputs 70% too often. It turns out that with FSRS-6 there may not be an "optimal" desired retention in terms of workload/knowledge, just "set DR as low as possible, and then even lower".


Expertium:
Jarrett wants to remove it because it outputs 70% too often. Me and another Anki contributor are trying something to solve it right now

Expertium:
With FSRS-5 there was only one "valley". With FSRS-6 there is still a valley, just somewhere far below 70%. So we're trying to modify what exactly CMRR is minimizing

PR #375, opened 2025-04-30
Update Resource Link

brishtibheja:
I'm moving the resource list in GDocs to forums. This is what @dae originally wanted so shouldn't be an issue.

Link: https://forums.ankiweb.net/t/collection-of-anki-resources/60044


Danika-Dakika:
Thanks!

PR #379, opened 2025-05-13
Document symlink behavior in styling section

adeluccar:
Users who rely on symlinks for custom CSS or font
files will find their files never make it up to
AnkiWeb and may be puzzled when they seem to work
on desktop.

Adding this note to the manual warns users up
front, prevents confusion, and should cut down on
future support questions related to styling.

### DIFF
--- a/src/media.md
+++ b/src/media.md
@@ -50,6 +50,11 @@ you should use Tools&gt;Check Media afterwards, to ensure the filenames are
 encoded correctly. If you skip this step, any filenames that are not compatible
 will be skipped when syncing.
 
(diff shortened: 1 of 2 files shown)


adeluccar:
This might trip up users on macOS systems using symlinks for their CSS and font files (happened to me). Hopefully character wrap settings are ok (used 80 chars per line).

dae:
I think this might be a better place: https://docs.ankiweb.net/media.html#manually-adding-media

dae:
Thank you!

PR #388, opened 2025-05-23
Include the "--locked" flag in the suggested "cargo install" invocation

bdesham:
I was unable to build the sync server (ankitects/anki#4031) until I included the `--locked` flag in the `cargo install` invocation. This seems to be the way the Dockerfile is built as well (as of the recent PR ankitects/anki#3856), so it seems like the manual should suggest using this flag too.

Since I was making changes in this area anyway, I bumped the example sync server version from 2.1.66 to 25.02.5.

### DIFF
--- a/src/sync-server.md
+++ b/src/sync-server.md
@@ -63,7 +63,7 @@ From Anki 2.1.66+, you can alternatively build a Rust implementation of the stan
 Make sure you have Rustup installed.
 
 ```
-cargo install --git https://github.com/ankitects/anki.git --tag 2.1.66 anki-sync-server
+cargo install --locked --git https://github.com/ankitects/anki.git --tag 2.1.66 anki-sync-server
 ```
 
 Replace 2.1.66 with whatever the latest Anki version is.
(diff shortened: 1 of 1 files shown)


src/sync-server.md
```diff
@@ -63,10 +63,10 @@ From Anki 2.1.66+, you can alternatively build a Rust implementation of the stan
 Make sure you have Rustup installed.
 
 ```
-cargo install --git https://github.com/ankitects/anki.git --tag 2.1.66 anki-sync-server
+cargo install --locked --git https://github.com/ankitects/anki.git --tag 25.02.5 anki-sync-server
 ```
 
-Replace 2.1.66 with whatever the latest Anki version is.
+Replace 25.02.5 with whatever the latest Anki version is.
```
Danika-Dakika, line 69: I know next to nothing about the sync server, so I probably shouldn't be the one to approve this -- but a couple notes: 

1. I believe it was `2.1.66` because that was the first version that supported this. (It matches the "From Anki 2.1.66+ ..." above.) Does that need to be updated to a more current version number?
2. Is `with whatever the latest Anki version is` correct? Or should a user match that command to the Anki version they actually have installed? [Genuine question -- I have no idea.]

PR #392, opened 2025-05-29
Include extra info from FAQs

brishtibheja:
Explained what happens to shared items upon data deletion.


Danika-Dakika:
Is this meant to be a complement to ankitects/faqs#47?

brishtibheja:
Yes, it is.

PR #402, opened 2025-06-03
document searching exact wording of a field

### LINKED ISSUE #5454: Document Searching All Fields
[forums](https://forums.ankiweb.net/t/potential-bug-newline-character-in-fields-html-disrupts-search-for-wildcard-fields/59617/2?u=sorata)

This might be useful to match exact wording used in a field. And it's simple enough that I don't see why it shouldn't be in docs.

brishtibheja:
Added `*:...` searches. I figured it can't put under "limiting to a field" so added to the end.

* Fix https://github.com/ankitects/anki/issues/5454

### DIFF
--- a/src/searching.md
+++ b/src/searching.md
@@ -480,3 +480,6 @@ interday learning cards due today.
 
 `prop:resched=0`\
 cards rescheduled today, either using **Set due date** or **Reschedule cards on change**.
+
+`*:dog`
+searches for notes with a field that exactly says "dog".


brishtibheja:
I mean we are not "limiting to a field" are we?

dae:
(please see the edits I made)

brishtibheja:
I think we should divide things by tasks that users would want to do, not by how Anki internally does it. I know "other searches" isn't helping in task-orientation but it's a better choice.

Danika-Dakika:
I think I'm missing something, because I'm struggling to understand the usefulness of this feature -- or why it needs to be documented at all. All searches already search all fields anyway, so is it that searching `*:dog` will return results where "dog" is the entire contents of the field, start to end -- and there's no other way to do that? Like a special case of: 

> `front:dog`
> finds notes with a Front field of exactly “dog”. A field that says “a dog” will not match.

If it's that, then I think it does belong in the "limiting to a field" section.

Issue #400, opened 2025-06-03
Add warning class where appropriate


brishtibheja:
In some parts of the docs, we use `<strong>` to make things stand out but a warning div can be more proper.

Some examples of where we can use it:

1. [Profile section](https://docs.ankiweb.net/profiles.html#profiles) warns user they cannot sync multiple profiles with one account.
2. The [security warning](https://docs.ankiweb.net/math.html#security-warning) for latex.

Refer to mdBook docs for how you do it: https://rust-lang.github.io/mdBook/format/mdbook.html#classwarning

brishtibheja:
These pages have FAQs at bottom: https://lazyguidejp.github.io/jp-lazy-guide/setupAnki/

PR #380, opened 2025-05-18
ADD info regarding limits (from faq)

GithubAnon0000:
Depends on ankitects/anki-manual#403.

This adds the info from the [FAQ section](https://faqs.ankiweb.net/are-there-limits-on-file-sizes-on-ankiweb.html) into the manual.

# Notable Changes
1. The incorrect info "250MB" has been updated to "300MB". See https://github.com/ankitects/faqs/pull/46.
2. I made the sections atomic.
3. I updated the "Tools>Media" part to adhere to the manuals style guide.
4. I reduced the amout of words on some sections.
5. I sometimes added the response / answer first (prepended "yes" or "no").
6. I used admonitions faq style, collapsed by default.

# Screenshots
<details><summary>Current entry in the FAQ:</summary>
<p>

anki

</p>
</details>

<details><summary>Entry in the docs, if this PR is merged:</summary>
<p>

anki

</p>
</details> 

<details><summary>[Old] Entry in the docs, if this PR is merged:</summary>
<p>

anki

</p>
</details> 

<hr>
Pinging @brishtibheja.

PR #403, opened 2025-06-03
ADD support for admonitions

GithubAnon0000:
Adds supports for admonitions in preparation for #400 and #380.

I followed https://github.com/tommilligan/mdbook-admonish?tab=readme-ov-file#installation for instructions. With this PR, the necessary css file has already been implemented, meaning an additional package is probably not necessary.


dae:
Please see my two extra commits for future reference.

PR #406, opened 2025-06-03
Add summary to background

GithubAnon0000:
Depends on #403.

I'm a huge fan of summaries. I chose "summary" instead of "abstract", because "abstract" felt too academic.

This PR introduces a very short summary, telling the readers the most important things of that section.

# Pictures

anki


dae:
It would be helpful if you could merge in the changes from the main branch so these PRs can be tested easily. Please also remove the space between the backticks and `admonish` - while it works with a space, it doesn't match the mdbook-admonish docs.

Deferring to @Danika-Dakika on this one

GithubAnon0000:
@abdnh, this one is ready.

PR #407, opened 2025-06-03
Modify backups (style guide and admonitions)

GithubAnon0000:
Depends on #403.

# Changes
1. This introduces new admonitions.
2. UI elements now are bold and use a "→".

# Images
anki
<hr>

anki


brishtibheja:
> Your message cut off.

I hate gmail...

The full message:

"I don't suggest using → as it's not a standard sign used in menu paths. The > sign is much more common and is being used in most parts of this docs."

> That's why I started using it everywhere and started to update everywhere.

Holy, and so we are trying to head towards different goals...

The arrow sign is a bad choice particularly because it's not a standard sign.

Plus, we're using it for procedural information. It would be better to use less complex signs like `>` so we can be sure that it wouldn't cause issues for screen reader users (screen readers might not read this).

GithubAnon0000:
@abdnh, this one is ready, unless we decide on using `->` instead of `→`. See daes suggestion above.

abdnh:
I'm also in favor of `>`. Will create an issue for mdbook-yapp.

GithubAnon0000:
@abdnh Okay, changed it. Should be good to go.

abdnh (approved):
Thank you!

PR #423, opened 2025-07-05
Added a brief explanation of CalVer

Expertium:
Let me know if there is a better place for this than Platform Notes


brishtibheja:
probably not a good place, we can instead note dates in the LP so the naming scheme is clearer (also the in-app about section)

Expertium:
What's LP?

brishtibheja:
landing page...

abdnh:
I think a good place is the Release Notes section in the new docs website: https://anki.mintlify.app/releases/betas/intro

It's not being updated at the moment but we plan to revive it in the future (probably after https://github.com/ankitects/anki/issues/5072 is done).

Created an issue to track it: https://github.com/ankitects/anki/issues/5476

PR #429, opened 2025-07-28
Add notes on the problem with the caddy reverse proxy read buffer size.

memchr:
The default HTTP read buffer size (4 KB) used by Caddy caused media downloads to fail with a stream reading error.

See https://forums.ankiweb.net/t/media-doesnt-sync-behind-caddy-error-303-when-downloading-media-upload-works-fine/60823


abdnh:
Thank you!

PR #432, opened 2025-08-02
Fix typo in searching.md

leedoughty:
Fixes typo in https://docs.ankiweb.net/searching.html#matching-special-characters


Danika-Dakika:
Great catch!

PR #441, opened 2025-08-25
Add removed notice for CMRR

Luc-Mcgrady:
ref: https://forums.ankiweb.net/t/cant-find-cmrr-option/66120/2


Expertium:
Would you please also remove mentions of CMRR from help tips in Anki?

src/deck-options.md
```diff
@@ -482,7 +482,11 @@ Use this option sparingly, as it will add a review entry to each of your cards,
 
 If you're first switching from SM-2 and still wish to use this option, we recommend you first create a backup, enable FSRS with rescheduling, and then if needed, you can undo or restore from the backup.
 
-### Compute Minimum Recommended Retention
+### Compute Minimum Recommended Retention (pre 25.07) {#compute-minimum-recommended-retention}
```
NamelessGO, line 485: Use <25.07 instead

avidseeker:
It'd be better to also add an explanation of *why* was it removed and what's the alternative?

Expertium:
@abdnh check this PR out as well

abdnh:
Thank you!

PR #448, opened 2025-09-09
Remove duplicate word in cloze explanation

leedoughty:
Removes duplicate word in https://docs.ankiweb.net/templates/errors.html#single-empty-cards


Danika-Dakika:
Great catch!

PR #451, opened 2025-09-19
[SyncServer] Add instructions for credentials

PetitPotiron:
It is **very** hard to understand, when self-hosting your syncserver, where you must put your credentials in order to synchronize the whole thing.

[This forum topic](https://forums.ankiweb.net/t/how-to-use-self-hosted-anki-server-on-linux-and-windows-anki-client/49264) illustrates the case. There's no prompt anywhere asking for the credentials, and it's *not* obvious that the credentials must be set in **AnkiWeb Account**, because it's written everywhere that's it's a different service, which is not affiliated with Anki...

I also struggled for a long time before finally figuring it out.

So anyways, the UI deserves to be enhanced as well.

In this PR,  I added instructions for where to put the user credentials when using a self-hosted syncserver.

### DIFF
--- a/src/sync-server.md
+++ b/src/sync-server.md
@@ -135,6 +135,7 @@ that the server binds to.
 
 ## Client Setup
 
+### Server URL
 You'll need to determine your computer's network IP address, and then
 point each of your Anki clients to the address, e.g something like
 `http://192.168.1.200:8080/`. The URL can be configured in the preferences.
@@ -148,6 +149,9 @@ Older desktop clients required you to define `SYNC_ENDPOINT` and
 `http://192.168.1.200:8080/sync/` and `http://192.168.1.200:8080/msync/`
 respectively. AnkiDroid clients before 2.16 require separate configuration for
 the two endpoints.
+
+### Credentials
+To be able to synchronize, put your user credential (`SYNC_USER1` for example) in the **AnkiWeb account** section of the preferences.
 
 ## Reverse Proxies
 
--- a/src/sync-server.md
+++ b/src/sync-server.md
@@ -151,7 +151,8 @@ respectively. AnkiDroid clients before 2.16 require separate configuration for
 the two endpoints.
 
 ### Credentials
-To be able to synchronize, put your user credential (`SYNC_USER1` for example) in the **AnkiWeb account** section of the preferences.
+To be able to synchronize, put your user credential (`SYNC_USER1` for example)
+in the **AnkiWeb account** section of the preferences.
 
 ## Reverse Proxies
 


PetitPotiron:
[This forum topic](https://forums.ankiweb.net/t/client-setup-section-in-sync-server-doc-needs-to-be-clarified/34863) also show that it's unclear

PetitPotiron:
> >    because it's written everywhere that's it's a different service, which is not affiliated with Anki...
>
> Not sure where you read that. AnkiWeb is an official webservice affiliated with Anki, see e.g. the official FAQs (https://faqs.ankiweb.net/anki-knockoffs.html).

I apologize for the misunderstanding, I don't know where I read that either.

> However, if you setup your custom sync server the Anki GUI still claims you sync with "AnkiWeb", even though that isn't the case anymore. You sync with your own sync server instead.

That's the problem, in my opinion. I believe it's not clear enough. And several forum posts show so too.

Even though it's written [at the beginning of the SyncServer doc](https://docs.ankiweb.net/sync-server.html#self-hosted-sync-server), not every user who uses a custom SyncServer will be its Administrator, and thus we should also explain it in [the section dedicated to the Client setup](https://docs.ankiweb.net/sync-server.html#client-setup), for both people who don't read beginnings of documents (almost always prerequisites, which can easily be fulfilled) and people who read the section they are interested in.

GithubAnon0000:
@PetitPotiron 

I also think it makes sense to include Credentials as you did in your PR in the client setup section. If the 80 column limit commit has been merged, then your PR is ready to merge in my opinion. Like this:

```suggestion
To be able to synchronize, put your user credential (`SYNC_USER1` for example)
in the **AnkiWeb account** section of the preferences.
```

(you can copy paste it. Github errors out with "internal error", so I cannot suggest a commit)

src/sync-server.md
```diff
@@ -149,6 +150,9 @@ Older desktop clients required you to define `SYNC_ENDPOINT` and
 respectively. AnkiDroid clients before 2.16 require separate configuration for
 the two endpoints.
 
+### Credentials
+To be able to synchronize, put your user credential (`SYNC_USER1` for example) in the **AnkiWeb account** section of the preferences.
```
GithubAnon0000:
```suggestion
To be able to synchronize, put your user credential (`SYNC_USER1` for example)
in the **AnkiWeb account** section of the preferences.
```

GithubAnon0000 (changes requested):
I can do the review again, the issue on githubs side is solved.

@PetitPotiron, please merge my suggestion. This PR LGTM after that's done.

src/sync-server.md
GithubAnon0000: @abdnh could you merge my suggestion here and then check if this looks good to you?

abdnh (approved):
Thank you!

abdnh:
Note: the docs are being moved to the main repo (we'll transfer open issues soon). I'll try to merge ready PRs, but future PRs should be submitted to https://github.com/ankitects/anki

GithubAnon0000:
Could we update the readme at the top to very prominently make this clear, please?

PR #449, opened 2025-09-16
Added "Help Me Decide", changed wording about DR

Expertium:
https://github.com/ankitects/anki-manual/pull/383 is inactive, so I made this
@Danika-Dakika feedback is welcome

PR #468, opened 2025-11-26
Remove "Evaluate FRSR Parameters" manual section

helmutcarter:
because that option has now been removed from Anki. Fixes issue ankitects/anki#5431


abdnh:
Looks like this was handled in #449. Thank you.

PR #472, opened 2025-12-26
fix(#471): replace legacy print.html button to prevent repeated print trigger

melvnl:
## Summary
This PR fixes the legacy print button behavior that causes the browser print dialog to be **re-triggered unintentionally** when navigating back or forward after canceling printing. See the issue https://github.com/ankitects/anki/issues/5427

The current implementation links to `/print.html`, which leads to poor print usage and UX on both desktop and mobile.

## Problem
- Clicking the print icon navigates to `/print.html`
- The browser print dialog opens
- Cancelling print and navigating back/forward retriggers the print dialog
- This creates a frustrating “loop-like” behavior

This is a UI glitch that **breaks expected print usage** and disrupts normal reading flow.

## Solution
- Intercept clicks on the legacy print link
- Prevent navigation to `/print.html`
- Invoke `window.print()` directly via JavaScript

This ensures:
- Print dialog opens once per explicit click
- No navigation occurs
- Back/forward navigation does not retrigger printing

## Implementation
- Added `js/print.js` with a click handler for the print button
- Included the script via `additional-js` in `book.toml`
- No theme overrides or build-toolchain changes

## Related Issue
https://github.com/ankitects/anki/issues/5427

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -20,6 +20,7 @@ assets_version = "3.0.3" # do not edit: managed by `mdbook-admonish install`
 git-repository-url = "https://github.com/ankitects/anki-manual/"
 cname = "docs.ankiweb.net"
 additional-css = ["css/misc.css", "css/mdbook-admonish.css"]
+additional-js = ["js/print.js"]
 smart-punctuation = true
 mathjax-support = true
 
--- a/js/print.js
+++ b/js/print.js
@@ -1,0 +1,46 @@
+(function() {
+    const isPrintPage = window.location.pathname.includes('print.html');
+    const urlParams = new URLSearchParams(window.location.search);
+    const hasTrigger = urlParams.get('do_print') === 'true';
+
+    // This only modifies the link on the current page to add the trigger.
+    const updatePrintButton = () => {
+        const printButton = document.querySelector('a[href*="print.html"]');
+        if (printButton && !printButton.href.includes('do_print')) {
+            const separator = printButton.href.includes('?') ? '&' : '?';
+            printButton.href += separator + 'do_print=true';
+        }
+    };
+
+    // --- PART 2: PRINT PAGE LOGIC ---
+    if (isPrintPage) {
+        // If we land on print.html without the trigger (e.g., via Back/Forward),
+        // use location.replace to bounce the user back to the previous chapter
+        // without adding a new entry to the history stack.
+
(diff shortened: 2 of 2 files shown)


GithubAnon0000 (changes requested):
Your proposal doesn't seem to work properly. It causes only the current page to be printed, whereas the current implementation prints the entire book.

Printing the current page only can already be easily achieved with the browser (e.g. pressing <kbd>Ctrl+p</kbd>), but printing the entire book is only possible with this button.

<hr>

In your issue you said this:
> The current Print button links to print.html (which doesn't exists)

That is incorrect, it does exist. After building the book you'll find a `book/` folder in the root, next to `book.toml`. The link to the entire book (which is being printed when the user clicks on that button), is `book/html/print.html`.

**Edit:** Also, please write "Fix #471" in your PRs description, so that github properly links this PR to that issue and auto closes #471 once this PR gets merged.

melvnl:
You’re absolutely right, sorry about that.
print.html does exist and is a special page generated by mdBook for printing the entire book. Thanks for the clarification.

After looking into it more closely, I realize my current proposal indeed only prints the current page, which doesn’t match the existing behavior. As you pointed out, printing a single page can already be done easily via the browser (e.g. Ctrl + P), while the value of this button is specifically printing the entire book.

I do have a small idea / gist on how we could tackle printing the whole book (not just the current page) while also fixing the re-triggered print function. I’ll take some time to refine this approach and update the PR accordingly.

Thanks again for the detailed explanation.

melvnl:
Hi, I've just updated the PR with new fix.

This fix addresses the issue where navigating back or forward to the print.html page causes the print dialog to re-trigger indefinitely. The current JavaScript-based solution uses a URL parameter to verify intent and history.replaceState to clean the history stack. While this avoids re-triggering, it can occasionally result in duplicate history entries (e.g., needing to click 'back' twice).

If a cleaner user experience is preferred, this can be resolved more robustly by overriding the index.hbs template to manage the print trigger via sessionStorage without affecting the URL or history flow

EDIT: I can also open a separate PR to demonstrate how the changes would look by modifying index.hbs to remove the automatic print behavior in print.html

abdnh:
Closing per https://github.com/ankitects/anki/issues/5427#issuecomment-5440665860

PR #473, opened 2025-12-30
FIX: put command into a codeblock

GithubAnon0000:
This is important for the styling but also might affected the font inself. E.g. I see this on the current site `‘anki –safemode’`. Only one `-` is incorrect though, causing the command to fail.


GithubAnon0000:
@abdnh, this one is ready.

PR #474, opened 2025-12-30
FIX: put command into code block

GithubAnon0000:
Makes it clear that `apt-get` is a command.


GithubAnon0000:
@abdnh, this one is ready.

PR #488, opened 2026-03-25
Update backups.md to include info about deleted.txt headers

hnvy:
Hi

As promised in ankitects/anki/pull/4454!

Harvey

### DIFF
--- a/src/backups.md
+++ b/src/backups.md
@@ -75,9 +75,21 @@ you can force a one-way sync in the preferences screen, or sync from a new devic
 
 ## Deletion log
 
-Anki logs deleted notes to a text file called deleted.txt in your
+Anki logs deleted notes to a text file called `deleted.txt` in your
 profile folder. These notes are in a text format that can be read by
-File&gt;Import, though please note the import feature only supports a
-single note type at one time, so if you have deleted notes from
-different note types, you'll need to split the file into separate files
-for each note type first.
+File&gt;Import. Historically, Anki only supported a single note type
+at one time. So, if you had deleted notes from different note types,
+you needed to split the file into separate files for each note type
+first.
+
+However, profiles created in Anki versions newer than 25.09.2 have the
+following snippet at the top of deleted.txt (hence allowing you to
+import `deleted.txt` without the need to split the file):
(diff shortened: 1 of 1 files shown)


src/backups.md
```diff
@@ -75,9 +75,21 @@ you can force a one-way sync in the preferences screen, or sync from a new devic
...
-single note type at one time, so if you have deleted notes from
-different note types, you'll need to split the file into separate files
-for each note type first.
+File&gt;Import. Historically, Anki only supported a single note type
+at one time. So, if you had deleted notes from different note types,
+you needed to split the file into separate files for each note type
+first.
+
+However, profiles created in Anki versions newer than 25.09.2 have the
+following snippet at the top of deleted.txt (hence allowing you to
```
GithubAnon0000:
```suggestion
following snippet at the top of `deleted.txt` (hence allowing you to
```

For consistency reasons, since you did this for the other mentions of that file as well.

src/backups.md
```diff
@@ -75,9 +75,21 @@ you can force a one-way sync in the preferences screen, or sync from a new devic
...
+import `deleted.txt` without the need to split the file):
+
+```
+#guid column:1
+#notetype column:2
+```
+
+Please note that `deleted.txt` is unique to each profile
+and is not interchangeable between different user profiles.
```
GithubAnon0000:
````suggestion
```admonish note
`deleted.txt` is unique to each profile and is not interchangeable between different user profiles.
```
````

There has been efforts into removing "Please" from the manual. We should thus remove it here, too.

Also, we could use admonish for the note part (but please test if the code actually renders a note block, I only wrote it by hand without testing it).

GithubAnon0000 (changes requested):

src/backups.md
hnvy: Seems to work ok [here](https://docs.ankiweb.net/backups.html#admonition-warning)

GithubAnon0000 (approved):
LGTM.

@abdnh, this one is ready, too.

abdnh:
Thank you both!

PR #497, opened 2026-06-11
Update for Anki 26.05

### LINKED ISSUE #4696: Update user documentation for new installer
Update the Anki user manual and download/install docs to explain the new
Briefcase-based installer, installation steps on each platform, and any
behavior differences from the previous UV-based launcher.

Docs site: https://docs.ankiweb.net/

Part of the Briefcase Installer initiative: #4629

abdnh:
Closes #490
Closes https://github.com/ankitects/anki/issues/4696

### DIFF
--- a/src/files.md
+++ b/src/files.md
@@ -62,13 +62,13 @@ modify the other files in the folder either.
 
 ## Program Files
 
-Anki's launcher is installed in the following locations by default:
+Anki is installed in the following locations by default:
 
 - Windows: `%LOCALAPPDATA%\Programs\Anki`
 - macOS: `/Applications/Anki.app`
 - Linux: `/usr/local/share/anki`
 
-When you install/update Anki with the launcher, it downloads support
+When you install/update Anki in versions 25.07 to 25.09.4, it downloads support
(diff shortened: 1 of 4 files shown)


fernandolins (commented):
@abdnh, the changes look good, but there are some places that need change, for instance: https://github.com/ankitects/anki-manual/pull/497/changes#diff-3488212f73f35a6db05b7be91cbe894d81af281dec073b9e0fe8b0fac035ff3eL28
Now Anki starts on 24.04, right? 

My point is: Should we focus only on the new installer? Or should we review the entire set of docs and update them?

abdnh:
I'm not sure about that particular issue in Ubuntu 24.04. It was reported in Anki 23.12.1 two years ago and probably no longer applies but I think it's fine to keep a link to the solution in case someone faces it.

> My point is: Should we focus only on the new installer? Or should we review the entire set of docs and update them?

I'm not aware of any other changes in 26.05 that require docs updates. Now that we have docs in the same repo after https://github.com/ankitects/anki/issues/4890, we can probably ensure docs are always up-to-date in the future.

PR #503, opened 2026-09-12
Document how to enable TTS on Linux

argrig666:
Linux users who import decks with native \`{{tts}}\` tags hit \`no players found for TTSTag(...)\` and often do not find the one-sentence add-on link in this chapter.

This change:

- States that Linux has no built-in TTS engine.
- Gives the exact Tools → Add-ons steps for the gTTS sample already linked here (\`391644525\`).
- Notes that \`{{tts lang:Field}}\` does not need a voice name, and that Apple/Microsoft voice names fall back by language.
- Mentions the sample add-on's profile-switch gap (player is registered at import only; see ankitects/anki-addons#40) and links a Linux player that also falls back to espeak-ng: https://github.com/argrig666/anki-linux-tts-player

Happy to drop the third-party GitHub link if you would rather wait until that add-on has an AnkiWeb code.


abdnh:
Please submit PRs to the [Anki repo](https://github.com/ankitects/anki/tree/main/docs-site) as noted in the README.

This repo will be archived soon.

